PR #25, opened 2020-09-23
Ftr:implement router to manage the API routing for the api gateway feature

### LINKED ISSUE #18: Router to route the requests
As a developer, I want to init a router to manage the requests that send to the proxy. The router should be able to find the target API, add a new target API, update and remove an existing API.
**Why is this needed**:

williamfeng323:
**What this PR does**:
Create router to store/find/update the apis

Fixes #18

```release-note

```


pkg/router/route.go
```diff
@@ -0,0 +1,160 @@
...
+	if _, ok := node.methods[method.HTTPVerb]; ok {
+		return errors.New(fmt.Sprintf("Method %s already exists in path %s", method.HTTPVerb, node.fullPath))
+	}
+	node.methods[method.HTTPVerb] = &method
+	return nil
+}
+
+func containParam(fullPath string) bool {
+	for _, s := range fullPath {
+		if s == ':' {
```
DogBaoBao, line 131: code `':' ` shoud be a const define.

pkg/router/route.go
```diff
@@ -0,0 +1,160 @@
...
+			return true
+		}
+	}
+	return false
+}
+
+func wildcardMatch(wildcardPath string, checkPath string) bool {
+	wildcardPath = strings.ToLower(wildcardPath)
+	checkPath = strings.ToLower(checkPath)
+	wPathSplit := strings.Split(wildcardPath[1:], "/")
```
DogBaoBao, line 141: like `':'`， I think `"/"` is a variable used `const`

pkg/config/api_config.go
```diff
@@ -62,14 +62,14 @@ type Resource struct {
 	Type        string     `yaml:"type"` // Restful, Dubbo
 	Path        string     `yaml:"path"`
 	Description string     `yaml:"description"`
-	Filters     []string   `yaml:"filters"`
```
DogBaoBao, line 65: maybe filed `Filters` can be here and always in `type Method struct` field too?

Issue #8, opened 2020-09-09
Add auto service discovery


cityiron:
**What would you like to be added**:
When config register info, will auto load from remote, change the service to a api(up/down) from rule.

**Why is this needed**:
When we not use dubbogo generic invoke, we need load invoke obj for self.

PR #26, opened 2020-09-23
auto file service discovery

zhangshen023:
**What this PR does**:
When config register info, will auto load from remote, change the service to a api(up/down) from rule.
#8 
Fixes #


pkg/api_load/api_manager.go
```diff
@@ -0,0 +1,230 @@
...
+		default:
+			time.Sleep(al.rateLimiterTime / 10)
+			break
+		}
+	}
+	return
+}
+
+// merge apis
+func (al *ApiManager) DoMergeApiTask() (skip bool, err error) {
```
williamfeng323, line 157: seems this function does not need to expose?

pkg/api_load/api_manager.go
```diff
@@ -0,0 +1,230 @@
...
+		// If apiLoadType is File,then try covering it's apis using other's apis from registry center
+		multiApisMerged := make(map[string]model.Api, 8)
+		var sortedApiLoader []int
+		sortedApiLoaderMap := make(map[int]ApiLoadType, len(al.ApiLoadTypeMap))
+		for apiLoadType, loader := range al.ApiLoadTypeMap {
+			sortedApiLoader = append(sortedApiLoader, loader.GetPrior())
+			sortedApiLoaderMap[loader.GetPrior()] = apiLoadType
+		}
+
+		sort.Ints(sortedApiLoader)
```
williamfeng323, line 176: instead of this sorting, will it be better to cerate a sortableApiLoaders type and  implement Len, Swap, Less interface to do the sorting?

zouyx (commented):
fix conflict and make the dog silent

PR #30, opened 2020-09-26
Ftr:implement router to manage the API routing for the api gateway feature

### LINKED ISSUE #18: Router to route the requests

williamfeng323:
**What this PR does**:
Create router to store/find/update the apis

Fixes #18

```release-note

```


DogBaoBao (approved):

pkg/config/api_config.go
```diff
@@ -69,7 +69,8 @@ type Resource struct {
 
 // Method defines the method of the api
 type Method struct {
-	OnAir              bool `yaml:"onAir"`
+	OnAir              bool     `yaml:"onAir"`
```
zouyx, line 72: what's the meaning of onair.......should you add some comment for this variable?

pkg/router/route.go
```diff
@@ -0,0 +1,161 @@
...
+// Node defines the single method of the router configured API
+type Node struct {
+	fullPath string
+	wildcard bool
+	methods  map[config.HTTPVerb]*config.Method
+}
+
+// Route defines the tree of router APIs
+type Route struct {
+	lock         sync.RWMutex
```
zouyx, line 45: Remember: if this object has a lock, dont copy this object, it will cause a lock copy problem.

Ref: https://github.com/apache/dubbo-go/issues/511

pkg/router/route.go
```diff
@@ -0,0 +1,161 @@
...
+	tree         *avltree.Tree
+	wildcardTree *avltree.Tree
+}
+
+// Put put a key val into the tree
+func (rt *Route) Put(fullPath string, method config.Method) error {
+	rt.lock.Lock()
+	defer rt.lock.Unlock()
+	fullPath = strings.ToLower(fullPath)
```
zouyx, line 54: I think you should not change the source.

pkg/router/route.go
```diff
@@ -0,0 +1,161 @@
...
+func (rt *Route) FindMethod(fullPath string, httpverb config.HTTPVerb) (*config.Method, bool) {
+	if n, found := rt.findNode(fullPath); found {
+		method, ok := n.methods[httpverb]
+		return method, ok
+	}
+	return nil, false
+}
+
+func (rt *Route) searchWildcard(fullPath string) (*Node, bool) {
+	wildcardPaths := rt.wildcardTree.Keys()
```
zouyx, line 112: add a read lock here?

pkg/router/route.go
```diff
@@ -0,0 +1,161 @@
...
+		}
+	}
+	return nil, false
+}
+
+func putMethod(node *Node, method config.Method) error {
+	if _, ok := node.methods[method.HTTPVerb]; ok {
+		return errors.New(fmt.Sprintf("Method %s already exists in path %s", method.HTTPVerb, node.fullPath))
+	}
+	node.methods[method.HTTPVerb] = &method
```
zouyx, line 126: `putMethod` should be a method of `Node`? and add a lock for this mehod?

pkg/router/route.go
```diff
@@ -0,0 +1,161 @@
...
+}
+
+func (rt *Route) findNode(fullPath string) (*Node, bool) {
+	fullPath = strings.ToLower(fullPath)
+	var n interface{}
+	var found bool
+	if n, found = rt.searchWildcard(fullPath); !found {
+		n, found = rt.tree.Get(fullPath)
+	}
+	return n.(*Node), found
```
zouyx, line 99: read lock?

pkg/router/route.go
```diff
@@ -0,0 +1,161 @@
...
+
+// UpdateMethod update the api method in the existing router node
+func (rt *Route) UpdateMethod(fullPath string, verb config.HTTPVerb, method config.Method) error {
+	rt.lock.Lock()
+	defer rt.lock.Unlock()
+	node, found := rt.findNode(fullPath)
+	if found {
+		if _, ok := node.methods[verb]; ok {
+			node.methods[verb] = &method
+		}
```
zouyx, line 87: move write to here?

pkg/router/route.go
```diff
@@ -0,0 +1,161 @@
...
+	tree         *avltree.Tree
+	wildcardTree *avltree.Tree
+}
+
+// Put put a key val into the tree
+func (rt *Route) Put(fullPath string, method config.Method) error {
+	rt.lock.Lock()
+	defer rt.lock.Unlock()
+	fullPath = strings.ToLower(fullPath)
+	wildcard := containParam(fullPath)
```
micln, line 55: move to line70 ?

pkg/router/route.go
```diff
@@ -0,0 +1,161 @@
...
+
+	if wildcardNode, found := rt.searchWildcard(fullPath); found {
+		return putMethod(wildcardNode, method)
+	}
+
+	node, ok := rt.tree.Get(fullPath)
+	if !ok {
+		ms := make(map[config.HTTPVerb]*config.Method)
```
micln, line 63: ms ?

pkg/router/route.go
micln, line 99: panic when n == nil

pkg/router/route.go
```diff
@@ -0,0 +1,161 @@
...
+			n, ok := rt.wildcardTree.Get(p)
+			return n.(*Node), ok
+		}
+	}
+	return nil, false
+}
+
+func putMethod(node *Node, method config.Method) error {
+	if _, ok := node.methods[method.HTTPVerb]; ok {
+		return errors.New(fmt.Sprintf("Method %s already exists in path %s", method.HTTPVerb, node.fullPath))
```
micln, line 124: errors.Errorf instead

pkg/router/route.go
micln, line 55: why not strings.Contains ?

pkg/router/route.go
```diff
@@ -0,0 +1,161 @@
...
+func containParam(fullPath string) bool {
+	for _, s := range fullPath {
+		if s == constant.PathParamIdentifier {
+			return true
+		}
+	}
+	return false
+}
+
+func wildcardMatch(wildcardPath string, checkPath string) bool {
```
micln, line 139: comment

pkg/common/constant/http.go
```diff
@@ -24,6 +24,9 @@ const (
 	HeaderValueJsonUtf8  = "application/json;charset=UTF-8"
 	HeaderValueTextPlain = "text/plain"
 	HeaderValueAll       = "*"
+
+	PathSlash           = "/"
+	PathParamIdentifier = ':'
```
micln, line 29: dose string is better ?

pkg/router/route.go
```diff
@@ -0,0 +1,161 @@
...
+	wildcardPath = strings.ToLower(wildcardPath)
+	checkPath = strings.ToLower(checkPath)
+	wPathSplit := strings.Split(wildcardPath[1:], constant.PathSlash)
```
micln, line 142: magic number

pkg/router/route.go
```diff
@@ -0,0 +1,161 @@
...
+	cPathSplit := strings.Split(checkPath[1:], constant.PathSlash)
+	if len(wPathSplit) != len(cPathSplit) {
+		return false
+	}
+	for i, s := range wPathSplit {
+		if containParam(s) {
+			cPathSplit[i] = s
+		}
```
micln, line 150: else 
if w[i] != c[i] return false

pkg/router/route.go
williamfeng323: done

pkg/router/route.go
williamfeng323: done

pkg/config/api_config.go
williamfeng323: done

pkg/router/route.go
williamfeng323: done

pkg/router/route.go
williamfeng323: done

pkg/router/route.go
williamfeng323: done

pkg/router/route.go
williamfeng323: fixed

pkg/router/route.go
williamfeng323: fixed

pkg/router/route.go
williamfeng323: fixed

pkg/router/route.go
williamfeng323: fixed

pkg/common/constant/http.go
williamfeng323: yup. fixed.

pkg/router/route.go
williamfeng323: fixed

pkg/router/route.go
williamfeng323: done

pkg/router/route.go
williamfeng323: done

pkg/router/route.go
```diff
@@ -0,0 +1,157 @@
...
+type Node struct {
+	fullPath string
+	wildcard bool
+	methods  map[config.HTTPVerb]*config.Method
+}
+
+// Route defines the tree of router APIs
+type Route struct {
+	lock         sync.RWMutex
+	tree         *avltree.Tree
```
zhangshen023, line 45: why do you choose avl-tree rather than rb-tree?

pkg/router/route.go
```diff
@@ -0,0 +1,157 @@
...
+		return errors.Errorf("Method %s already exists in path %s", method.HTTPVerb, node.fullPath)
+	}
+	node.methods[method.HTTPVerb] = &method
+	return nil
+}
+
+// wildcardMatch validate if the checkPath meets the wildcardPath,
+// for example /vought/12345 should match wildcard path /vought/:id;
+// /vought/1234abcd/status should not match /vought/:id;
+func wildcardMatch(wildcardPath string, checkPath string) bool {
```
zhangshen023, line 133: what if the wildcard like "/a/a-:id/"?

pkg/router/route.go
williamfeng323: I think in our use case, the tree is relatively stable and will not change frequently; Moreover, the performance of searching the tree is more critical, so I choose to use the avl-tree.

pkg/router/route.go
williamfeng323: good question. In current logic, it will be equal to /a/:id, and the prefix a- will be ignored. I think we'd better forbid the /a/a-:id to be used. What do you think?

zhangshen023 (approved):

AlexStocks (approved):

PR #58, opened 2020-11-17
Filter move

cityiron:
**What this PR does**:
Filter move to self package and create responseFilter

Fixes #

```release-note

```


yqxu (approved):

pkg/common/constant/proxy.go
```diff
@@ -19,13 +19,27 @@ package constant
...
 
-// body passthrough
+// body passthrough.
 const (
 	DefaultBodyAll = "_all"
 )
+
+// proxy code.
+const (
+	TimeoutError    = "DGP1005"
```
williamfeng323, line 35: Do we need the const error message or we can use 408 status code?

pkg/filter/response/response.go
```diff
@@ -15,22 +15,75 @@
...
+	contexthttp "github.com/dubbogo/dubbo-go-proxy/pkg/context/http"
+	"github.com/dubbogo/dubbo-go-proxy/pkg/filter"
 )
 
-// NewDubboResponse create dubbo response
-func NewDubboResponse(data interface{}) *client.Response {
-	r, err := dealResp(data, true)
+func init() {
+	strategy := os.Getenv(constant.EnvResponseStrategy)
+	if strategy == "" {
```
williamfeng323, line 40: use len(strategy) != 0?

pkg/common/constant/proxy.go
cityiron: agree

pkg/filter/response/response.go
cityiron: yes, better

williamfeng323 (approved):

zouyx (approved):

PR #66, opened 2020-12-03
add uri support for http

### LINKED ISSUE #56: URI mapping
retrieve value from URI. For example, giving API config test.com/user/:id, when calling test.com/user/12345, we should be able to retrieve 12345 from id by configuring mapping name uri.id
**Why is this needed**:

williamfeng323:
**What this PR does**:
Add support to mapping URI to upstream service.
1. If the inbound request and the backend http service contains wildcard params, it can automatically map it.
2. Config the mapping param uri.xx to map to any params you need.

Fixes #56

```release-note

```

### DIFF
--- a/pkg/client/http/http.go
+++ b/pkg/client/http/http.go
@@ -27,7 +27,7 @@ import (
 )
 
 import (
-	"github.com/apache/dubbo-go/common/constant"
+	dgConstant "github.com/apache/dubbo-go/common/constant"
 	dg "github.com/apache/dubbo-go/config"
 	"github.com/apache/dubbo-go/protocol/dubbo"
 	"github.com/pkg/errors"
@@ -35,6 +35,7 @@ import (
 
 import (
 	"github.com/dubbogo/dubbo-go-proxy/pkg/client"
+	"github.com/dubbogo/dubbo-go-proxy/pkg/common/constant"
 )
 
 // RestMetadata dubbo metadata, api config
@@ -97,8 +98,6 @@ func (dc *Client) Close() error {
 
 // Call invoke service
 func (dc *Client) Call(req *client.Request) (resp interface{}, err error) {
-	urlStr := req.GetURL()
-	u, err := url.ParseRequestURI(urlStr)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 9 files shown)


pkg/router/api.go
```diff
@@ -19,11 +19,12 @@ package router
 
 import (
 	"net/url"
-	"strings"
 )
 
 import (
+	"github.com/dubbogo/dubbo-go-proxy/pkg/common/constant"
 	"github.com/dubbogo/dubbo-go-proxy/pkg/config"
+	"strings"
```
cityiron, line 27: split

cityiron (changes requested):

cityiron (approved):

pkg/client/http/http.go
```diff
@@ -112,10 +111,13 @@ func (dc *Client) Call(req *client.Request) (resp interface{}, err error) {
...
 
-	newReq, err := http.NewRequest(req.IngressRequest.Method, urlStr, params.Body)
+	targetURL, err := dc.parseURL(req, *params)
+	if err != nil {
+		return nil, err
+	}
+
+	newReq, err := http.NewRequest(req.IngressRequest.Method, targetURL, params.Body)
 
 	httpClient := &http.Client{Timeout: 5 * time.Second}
```
zhangshen023, line 122: a little tip: move this timeout option to dc

pkg/client/http/mapper.go
```diff
@@ -207,3 +207,19 @@ func setMapWithPath(targetMap map[string]interface{}, path string, val interface
...
+	_, fromKey, err := client.ParseMapSource(mp.Name)
+	if err != nil {
+		return nil, nil, "", nil, err
+	}
+	to, toKey, err := client.ParseMapSource(mp.MapTo)
+	if err != nil {
+		return nil, nil, "", nil, err
+	}
+	return target, fromKey, to, toKey, nil
+}
```
zhangshen023, line 225: if returning params are too much, `
func funcname()(param1 string,param2 int){
    return
}
` may be clean and better

zhangshen023: and more comments?

pkg/client/http/mapper.go
```diff
@@ -42,35 +42,30 @@ var mappers = map[string]client.ParamMapper{
...
+		Query:     url.Values{},
+		Body:      ioutil.NopCloser(bytes.NewReader([]byte(""))),
+		URIParams: url.Values{},
 	}
 }
 
 type queryStringsMapper struct{}
 
 // nolint
 func (qs queryStringsMapper) Map(mp config.MappingParam, c *client.Request, rawTarget interface{}, option client.IOption) error {
```
zhangshen023, line 67: Can "qs" is defined as pointer?

zhangshen023 (commented):
@williamfeng323 thanks for contributing.

pkg/client/http/http.go
williamfeng323: I think It should be a configuration read from the config file or get the default value.

pkg/client/http/mapper.go
williamfeng323: because it will not use anything within the queryStringsMapper struct, and it is a lite struct. So I think it might be better to make it a non-pointer function

zhangshen023 (approved):

yqxu (approved):

PR #75, opened 2020-12-12
Http sample

williamfeng323:
**What this PR does**:
1. Fixed a bug in HTTP client that it might get the previous error request's response as the current request's response.
2. Add sample for HTTP-to-HTTP, including the uri parameters.
3. There is a bug in replacepath filter. In the filter, it will replace the original URL path with the configured target path. It makes the URI parameters value missing. So before this bug fixed, the URI analysis will not be able to use(for HTTP-to-http proxy).

Fixes #

```release-note

```

### DIFF
--- a/sample/http/server/app/main.go
+++ b/sample/http/server/app/main.go
@@ -28,8 +28,6 @@ import (
 
 import (
 	"github.com/dubbogo/dubbo-go-proxy/pkg/common/constant"
-	"math/rand"
-	"strings"
 )
 
 func main() {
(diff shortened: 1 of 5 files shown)


sample/http/server/app/main.go
```diff
@@ -26,10 +26,12 @@ import (
 
 import (
 	"github.com/dubbogo/dubbo-go-proxy/pkg/common/constant"
+	"math/rand"
```
cityiron, line 29: split

cityiron (approved):

xiaoliu10 (approved):

zhangshen023 (approved):

PR #85, opened 2020-12-22
Fix registry url reference type

### LINKED ISSUE #83: modify registry url use reference type
 make registry url use reference type

**Why is this needed**:

xiaoliu10:
**What this PR does**:
 upgrade  dubbo go v1.5.5-rc1 than common.URL use reference type
so modify dubbo go proxy url value type 

Fixes #83

```release-note

```


cityiron (approved):

pkg/registry/consul.go
```diff
@@ -70,7 +70,7 @@ func (crl *ConsulRegistryLoad) GetCluster() (string, error) {
 	return crl.cluster, nil
 }
 
-func (crl *ConsulRegistryLoad) transfer2Url(service consul.AgentService) (common.URL, error) {
+func (crl *ConsulRegistryLoad) transfer2Url(service consul.AgentService) (*common.URL, error) {
```
zhangshen023, line 73: IMO, less pointer may be better

xiaoliu10: in dubbo go v1.5.5-rc1,  common.Url was chanaged to a reference type, and there is a lock in the url , If I evaluate it (*URL), I get an error, can not pass linter

zhangshen023: OK

zhangshen023 (approved):

AlexStocks (approved):

PR #159, opened 2021-04-24
Refractor pixiu_start.go

lmhmhl:
Here are some changes as follow:
1. Refractor `pixiu_start.go`, which extract the inner method as an independent method and delete some useless blank line and comment.
2. Change the plugin registered by hand to auto using `init()` method when import the related package.


cmd/pixiu/control.go
```diff
@@ -18,6 +18,7 @@
 package main
 
 import (
+	"github.com/apache/dubbo-go-pixiu/pkg/model"
```
AlexStocks, line 21: move it to the 3rd import block.

cmd/pixiu/control.go
```diff
@@ -133,3 +97,47 @@ var (
...
+	logger.InitLog(logConfPath)
+	if err := GetAPIConfig(bootstrap, apiConfigPath); err != nil {
+		return err
+	}
+	InitNumCPU(c)
+	pixiu.Start(bootstrap)
+	return nil
+}
+
+func GetAPIConfig(bootstrap *model.Bootstrap, apiConfigPath string) error {
```
AlexStocks, line 119: func GetAPIConfig(bootstrap *model.Bootstrap, apiConfigPath string) error {
	metaConfig := bootstrap.GetAPIMetaConfig()
	if metaConfig == nil {
		if _, err := config.LoadAPIConfigFromFile(apiConfigPath); err != nil {
			logger.Errorf("load api config error:%+v", err)
			return err
		}
                 return nil
         }

	if _, err := config.LoadAPIConfig(bootstrap.GetAPIMetaConfig()); err != nil {
		logger.Errorf("load api config from etcd error:%+v", err)
		return err
	}
	return nil
}

xiaoliu10 (approved):

williamfeng323 (approved):

williamfeng323:
review dog的麻烦处理下

mark4z:
this pr maybe can not merged, I will refractor the filter later. @lmhmhl @xiaoliu10

PR #166, opened 2021-05-12
fix version field

### LINKED ISSUE #138: Incorrect version field
**What happened**:
image
pixiu is now v0.2.0 version, but the version field is still 0.1.0

**What you expected to happen**:

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

lbh2001:
**What this PR does**:
Fixed version field.

Fixes #138 

```release-note
NONE
```

### DIFF
--- a/cmd/pixiu/pixiu.go
+++ b/cmd/pixiu/pixiu.go
@@ -40,7 +40,7 @@ import (
 )
 
 // Version pixiu version
-var Version = "0.1.0"
+var Version = "0.3.0"
 
 // main pixiu run method
 func main() {


AlexStocks (approved):

cmd/pixiu/pixiu.go
```diff
@@ -40,7 +40,7 @@ import (
 )
 
 // Version pixiu version
-var Version = "0.1.0"
+var Version = "0.2.0"
```
AlexStocks, line 43: 0.3.0?

AlexStocks (approved):

cmd/pixiu/pixiu.go
xiaoliu10: maybe v0.2.1 can merge this PR

xiaoliu10 (approved):

cmd/pixiu/pixiu.go
lbh2001: @xiaoliu10  Sorry for my bothering, but I've sent something about [ospp](https://summer.iscas.ac.cn/#/org/orgdetail/dubbogo) program to your email, and there is also an issue I've created about that. Could you please take a look? Thank you so much! :blush:

kezhenxu94:
I'd rather extract the variable as a build argument, otherwise, this kind of issue is highly possible to occur next time

lbh2001:
> I'd rather extract the variable as a build argument, otherwise, this kind of issue is highly possible to occur next time

So we should put variables like this into config file in the future, right? Should I fix it?

kezhenxu94:
> > I'd rather extract the variable as a build argument, otherwise, this kind of issue is highly possible to occur next time
> 
> So we should put variables like this into config file in the future, right? Should I fix it?

What I mean is a build argument, not configuration file, take the below as an example

https://github.com/apache/skywalking-cli/blob/d06d1ebf565756885650061ddd2e1fe527417052/cmd/swctl/main.go#L43

https://github.com/apache/skywalking-cli/blob/d06d1ebf565756885650061ddd2e1fe527417052/Makefile#L18

https://github.com/apache/skywalking-cli/blob/d06d1ebf565756885650061ddd2e1fe527417052/Makefile#L40

https://github.com/apache/skywalking-cli/blob/d06d1ebf565756885650061ddd2e1fe527417052/Makefile#L71

kezhenxu94:
A release manager should take care of the build argument when drafting a new release, and that doesn't require code changes

lbh2001:
> A release manager should take care of the build argument when drafting a new release, and that doesn't require code changes

I've understood what you mean, and I will try my best to implement it after learning about Makefile. But before that, I wonder why one of the checks didn't pass and how to fix it. (Maybe this question is ridiculous :cry:)

kezhenxu94:
> Maybe this question is ridiculous

It's totally ok, there is no such question as ridiculous in community. Don't be panic

PR #171, opened 2021-05-18
make Pixiu Admin config management finer-grained

ztelur:
### What this PR does:

- init from api_config fetched from etcd remotely
- wath api_config change event from etcd
- use gost etcd and zk dependency


pkg/config/api_config.go
```diff
@@ -67,64 +80,116 @@ func LoadAPIConfigFromFile(path string) (*fc.APIConfig, error) {
...
 	}
+	// TODO: 其他需要从远端获取配置的初始化操作
 
+	go listenResourceAndMethodEvent(metaConfig.APIConfigPath)
+	// TODO: 其他监控配置的操作，比如 PluginGroup 等
 	return apiConfig, nil
 }
 
-func initAPIConfigFromString(content string) error {
+func initAPIConfigFromKVList(kList, vList []string) error {
```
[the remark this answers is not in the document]
ztelur: resource and method data store in etcd as parent path relation. with this design, it' a little difficulty to split this function when fetch data from etcd use prefix=true option

pkg/config/api_config.go
```diff
@@ -67,64 +80,116 @@ func LoadAPIConfigFromFile(path string) (*fc.APIConfig, error) {
 
 // LoadAPIConfig load the api config from config center
 func LoadAPIConfig(metaConfig *model.APIMetaConfig) (*fc.APIConfig, error) {
-	client = etcdv3.NewConfigClient(
+	client, _ = etcdv3.NewConfigClientWithErr(
```
williamfeng323, line 83: should we handle the error return from the new client function?

williamfeng323: It also limited the config center can only be etcd. Could we get the config center client from something like getConfigCenterClient(some_indicator)? It is not generic enough I think.

pkg/config/api_config.go
```diff
@@ -67,64 +80,116 @@ func LoadAPIConfigFromFile(path string) (*fc.APIConfig, error) {
...
+
+	for i, k := range kList {
+		v := vList[i]
 
-		valid := validateAPIConfig(apiConf)
-		if !valid {
-			return perrors.Errorf("api config not valid error %v", perrors.WithStack(err))
+		// handle resource
+		re := getCheckResourceRegexp()
+		if m := re.Match([]byte(k)); m {
```
williamfeng323, line 114: maybe we could use below judgment to reduce the complexity.
```
if m := re.Match([]byte(k)); m {
    continue
}
```

pkg/config/api_config.go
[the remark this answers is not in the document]
williamfeng323: I agree with cityiron. the complexity of this function is high. If we introduce sonarqube, it cannot pass the check.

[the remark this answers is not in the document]
ztelur: ok, I will make handle resource and handle method two method

pkg/config/api_config.go
ztelur: already handle the error return from the new client function

mark4z (approved):

pkg/registry/zookeeper.go
```diff
@@ -17,14 +17,14 @@
 package registry
 
 import (
+	gxzookeeper "github.com/dubbogo/gost/database/kv/zk"
```
AlexStocks, line 20: move it to the 2rd import block.

pkg/registry/zookeeper_test.go
```diff
@@ -18,6 +18,7 @@ package registry
 
 import (
 	"fmt"
+	gxzookeeper "github.com/dubbogo/gost/database/kv/zk"
```
AlexStocks, line 21: move it to 2rd import block

williamfeng323 (approved):

PR #190, opened 2021-06-15
Fix timeout config overridden

### LINKED ISSUE #189: If we configure more than one clusters in conf.yaml, the timeout field of former cluster will be overridden
**What happened**:

If we configure more than one cluster in `conf.yaml`, the timeout field (connect_timeout and request_timeout) of the former cluster will be overridden.

**Solution**:

We can move the timeout config outside the clusters.

alchemy-lee:
**What this PR does**:

This PR moves the timeout config outside the clusters. In detail, it moves `connect_timeout` and `request_timeout` from `clusters` into `timeout_config` in ` conf.yaml`.

Fixes #189


xiaoliu10 (approved):

AlexStocks (approved):

pkg/config/config_load_test.go
```diff
@@ -134,6 +132,10 @@ func TestMain(m *testing.M) {
 					},
 				},
 			},
+			TimeoutConfig: model.TimeoutConfig{
+				ConnectTimeoutStr: "5s",
+				RequestTimeoutStr: "10s",
+			},
```
williamfeng323, line 138: can we add 1-2 tests to ensure the timeout config works?

williamfeng323 (approved):

PR #341, opened 2022-01-09
add benchmark test case for performance test

yqxu:
**What this PR does**:
add benchmark test case for performance test


mark4z:
Resolve confilcts pls.

PR #422, opened 2022-05-11
fix: modify tracing filter config method.

maxingg:
**What this PR does**:

Update TraceFilterFactory's Config method, this should not return nil. In addition, resolve in samples doesn't need a tracing config.

### DIFF
--- a/pkg/filter/tracing/tracing.go
+++ b/pkg/filter/tracing/tracing.go
@@ -29,7 +29,7 @@ import (
 	"github.com/apache/dubbo-go-pixiu/pkg/tracing"
 )
 
-const jaegerTraceIDInHeader = "uber-trace-id"
+const TraceIDInHeader = "uber-trace-id"
 
 // nolint
 func init() {
@@ -47,6 +47,8 @@ type (
 	TraceFilter struct {
 		span trace.Span
 	}
+
+	Config struct{}
 )
 
 func (ap *Plugin) Kind() string {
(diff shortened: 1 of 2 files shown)


pkg/filter/tracing/tracing.go
```diff
@@ -29,7 +29,7 @@ import (
 	"github.com/apache/dubbo-go-pixiu/pkg/tracing"
 )
 
-const jaegerTraceIDInHeader = "uber-trace-id"
+const TraceIDInHeader = "uber-trace-id"
```
AlexStocks: change the value to "dubbo-go-trace-id"?

maxingg: "uber-trace-id" is used by jaeger. Since we can support multiple exporters, "dubbo-go-trace-id" is better actually.

AlexStocks: so, pls change it.

pkg/filter/tracing/tracing.go
```diff
@@ -29,7 +29,7 @@ import (
 	"github.com/apache/dubbo-go-pixiu/pkg/tracing"
 )
 
-const jaegerTraceIDInHeader = "uber-trace-id"
+const TraceIDInHeader = "dubbo-go-trace-id"
```
mark4z: dubbo-go-pixiu-trace-id better

ztelur (approved):

PhilYue (approved):

AlexStocks (approved):

PR #668, opened 2025-04-20
feat: supports tcp/http/https for health check & fix: Fix ignoring the domains field #667

### LINKED ISSUE #667: domain specified in endpoint in conf.yaml will be ignored
**What happened**:

If you specify a domain name for the endpoint, it will be ignored. You can only specify an IP address. 

In this conf.yaml, the domain I specified in line 57, ```"api.deepseek.com"``` will be ignored, pixiu will use default ip_address and default port as the target endpoint, i.e. ```0.0.0.0:8081```.

```yaml
static_resources:
  listeners:
    - name: "llm_proxy"
      protocol_type: "HTTP"
      address:
        socket_address:
          address: "0.0.0.0"
          port: 8888
      filter_chains:
          filters:
            - name: dgp.filter.httpconnectionmanager
              config:
                route_config:
                  routes:
                    - match:
                        prefix: "/chat/completions"
                      route:
                        cluster: "chat"
                        cluster_not_found_response_code: 505
                http_filters:
                  - name: dgp.filter.http.httpproxy
                    config:
                      maxIdleConns: 100
                      maxIdleConnsPerHost: 100
                      MaxConnsPerHost: 100
                  - name: dgp.filter.llm.tokenizer
      config:
        idle_timeout: 5s
        read_timeout: 5s
        write_timeout: 5s
  clusters:
    - name: "chat"
      lb_policy: "lb"
      endpoints:
        - id: 1
          socket_address:
            domains:
              - "api.deepseek.com"
  shutdown_config:
    timeout: "60s"
    step_timeout: "10s"
    reject_policy: "immediacy"
```

For example, in the httpfilter ```pkg/filter/http/httpproxy/routerfilter.go```, the endpoint address is obtained through the ```GetAddress``` function.

```go
	parsedURL := url.URL{
		Host:     endpoint.Address.GetAddress(),
		Scheme:   "https",
		Path:     r.URL.Path,
		RawQuery: r.URL.RawQuery,
	}
```

The source code of ```GetAddress``` in the second line which is:

```pkg/model/base.go```
```go
	SocketAddress struct {
		Address      string   `default:"0.0.0.0" yaml:"address" json:"address" mapstructure:"address"`
		Port         int      `default:"8881" yaml:"port" json:"port" mapstructure:"port"`
		ResolverName string   `yaml:"resolver_name" json:"resolver_name" mapstructure:"resolver_name"`
		Domains      []string `yaml:"domains" json:"domains" mapstructure:"domains"`
		CertsDir     string   `yaml:"certs_dir" json:"certs_dir" mapstructure:"certs_dir"`
	}

func (a SocketAddress) GetAddress() string {
	return fmt.Sprintf("%s:%v", a.Address, a.Port)
}
```

The code ignores ```domian``` field, which takes default ```Address``` and default ```port```

**What you expected to happen**:

The ```domain``` field should not be ignored, in the httpfilter above, ```parsedURL.host``` should be the domain denoted in conf.yaml instead of the default ```Address``` and default ```port```.

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

Alanxtl:
**What this PR does**:

- feat: Now can specify the protocol field in the conf file as the target protocol to be checked by health_checks, previewsly protocol only supports for tcp and the ```protocol``` field is neglected. Currently only supports tcp/http/https, for example:

```yaml
health_checks:
  - protocol: "https"
    timeout: 1s
    interval: 2s
    healthy_threshold: 4
    unhealthy_threshold: 4
```

The new implmented health check also support specifying the domains field.

- fix #667: now can specify the domains field in the conf file as endpoint address.

### DIFF
--- a/pkg/cluster/healthcheck/tcp.go
+++ b/pkg/cluster/healthcheck/tcp.go
@@ -18,12 +18,7 @@
 package healthcheck
 
 import (
-	"net"
 	"time"
-)
-
-import (
-	"github.com/apache/dubbo-go-pixiu/pkg/logger"
 )
 
 type TCPChecker struct {
@@ -32,13 +27,7 @@ type TCPChecker struct {
 }
 
 func (s *TCPChecker) CheckHealth() bool {
-	conn, err := net.DialTimeout("tcp", s.address, s.timeout)
-	if err != nil {
-		logger.Infof("[health check] tcp checker for host %s error: %v", s.address, err)
-		return false
-	}
-	conn.Close()
-	return true
+	return TcpConn(s.address, 8081, s.timeout)
 }
 
 func (s *TCPChecker) OnTimeout() {}
--- a/pkg/cluster/healthcheck/tcp.go
+++ b/pkg/cluster/healthcheck/tcp.go
@@ -27,7 +27,7 @@ type TCPChecker struct {
 }
 
 func (s *TCPChecker) CheckHealth() bool {
-	return TcpConn(s.address, 8081, s.timeout)
+	return TcpConn(s.address, "", s.timeout)
 }
 
 func (s *TCPChecker) OnTimeout() {}
--- a/pkg/cluster/healthcheck/tcp.go
+++ b/pkg/cluster/healthcheck/tcp.go
@@ -27,7 +27,7 @@ type TCPChecker struct {
 }
 
 func (s *TCPChecker) CheckHealth() bool {
-	return TcpConn(s.address, "", s.timeout)
+	return CheckTcpConn(s.address, "", s.timeout)
 }
 
 func (s *TCPChecker) OnTimeout() {}
--- a/pkg/cluster/healthcheck/utils.go
+++ b/pkg/cluster/healthcheck/utils.go
@@ -1,3 +1,20 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
 package healthcheck
 
 import (
--- a/pkg/cluster/healthcheck/utils.go
+++ b/pkg/cluster/healthcheck/utils.go
@@ -1,0 +1,27 @@
+package healthcheck
+
+import (
+	"github.com/apache/dubbo-go-pixiu/pkg/logger"
+	"net"
+	"strconv"
+	"strings"
+	"time"
+)
+
+func TcpConn(tarAddr string, port int, timeout time.Duration) bool {
+	if _, _, err := net.SplitHostPort(tarAddr); err != nil {
+		if strings.Contains(err.Error(), "missing port in address") {
+			tarAddr = tarAddr + ":" + strconv.Itoa(port)
+		} else {
+			logger.Infof("[health check] invalid address format: %s", tarAddr)
+			return false
+		}
+	}
+	conn, err := net.DialTimeout("tcp", tarAddr, timeout)
+	if err != nil {
+		logger.Infof("[health check] http checker for host %s error: %v", tarAddr, err)
+		return false
+	}
+	conn.Close()
+	return true
+}
--- a/pkg/cluster/healthcheck/utils.go
+++ b/pkg/cluster/healthcheck/utils.go
@@ -27,37 +27,37 @@ import (
 	"github.com/apache/dubbo-go-pixiu/pkg/logger"
 )
 
-func TcpConn(tarAddr string, tarPort string, timeout time.Duration) bool {
+func TcpConn(address string, port string, timeout time.Duration) bool {
 
-	if tarPort == "" {
-		// if tarPort is empty, tarAddr must has port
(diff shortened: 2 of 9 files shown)


pkg/cluster/healthcheck/utils.go
```diff
@@ -20,25 +20,43 @@ package healthcheck
 import (
 	"github.com/apache/dubbo-go-pixiu/pkg/logger"
```
AlexStocks: 重新 format 下

Alanxtl: done

pkg/cluster/healthcheck/tcp.go
```diff
@@ -14,30 +14,20 @@
...
 	"time"
 )
 
-import (
-	"github.com/apache/dubbo-go-pixiu/pkg/logger"
-)
-
 type TCPChecker struct {
-	addr    string
+	address string
```
AlexStocks, line 25: 改这个变量名字有啥意义呢?addr也能望文生义呀

pkg/cluster/healthcheck/utils.go
```diff
@@ -0,0 +1,65 @@
...
+	"net"
+	"strings"
+	"time"
+)
+
+import (
+	"github.com/apache/dubbo-go-pixiu/pkg/logger"
+)
+
+func TcpConn(tarAddr string, tarPort string, timeout time.Duration) bool {
```
AlexStocks: 前缀 tar 嘛意思？

pkg/cluster/healthcheck/tcp.go
Alanxtl: 在```pkg/model/base.go```里面和其他model里面定义的ip地址也都是使用address，我想的是和model里面定义的保持一致

pkg/cluster/healthcheck/utils.go
Alanxtl: target

AlexStocks: 那就去掉这个前缀吧，无意义

Alanxtl: done

pkg/cluster/healthcheck/utils.go
```diff
@@ -0,0 +1,65 @@
...
+	"net"
+	"strings"
+	"time"
+)
+
+import (
+	"github.com/apache/dubbo-go-pixiu/pkg/logger"
+)
+
+func TcpConn(address string, port string, timeout time.Duration) bool {
```
No-SilverBullet: 这里的命名不好，返回bool类型，是个检查连接的动作，但是名字叫TcpConn，CheckXXX这种会更清楚一点，

Alanxtl: done, 改成了CheckTcpConn

pkg/cluster/healthcheck/utils.go
```diff
@@ -0,0 +1,65 @@
...
+	"net"
+	"strings"
+	"time"
+)
+
+import (
+	"github.com/apache/dubbo-go-pixiu/pkg/logger"
+)
+
+func CheckTcpConn(address string, port string, timeout time.Duration) bool {
```
AlexStocks, line 30:
```go
func CheckTcpConn(address, port string, timeout time.Duration) error {
	normalizedAddress, err := normalizeAddress(address, port)
	if err != nil {
		return fmt.Errorf("[health check] address error: %w", err)
	}

	conn, err := net.DialTimeout("tcp", normalizedAddress, timeout)
	if err != nil {
		return fmt.Errorf("[health check] TCP connection to %s failed: %w", normalizedAddress, err)
	}
	defer conn.Close()
	return nil
}

func normalizeAddress(address, port string) (string, error) {
	host, parsedPort, err := net.SplitHostPort(address)
	if err == nil {
		if port != "" && parsedPort != port {
			return net.JoinHostPort(host, port), nil
		}
		return address, nil
	}

	// 解析失败，可能是因为缺少端口
	if port == "" {
		return "", fmt.Errorf("no port specified and address is missing port: %s", address)
	}

	// 如果是缺少端口的错误
	if strings.Contains(err.Error(), "missing port in address") {
		return net.JoinHostPort(address, port), nil
	}

	return "", fmt.Errorf("invalid address format: %s", address)
}
```

你这个函数写的比较绕，我用个 gpt 重构了下，你看着改进下，新函数还支持 ipv6

测试代码1：
err := CheckTcpConn("qq.com", "80", 3*time.Second)
if err != nil {
	logger.Warnf("TCP check failed: %v", err)
} else {
	logger.Infof("TCP check succeeded")
}

测试代码2：
err := CheckTcpConn("[::1]", "8080", 2*time.Second)

可以把测试代码改成单测

Alanxtl: 单测已添加，但是这个gpt重构的逻辑不太合理，我改成这样了。

```go
func CheckTcpConn(address string, port string, timeout time.Duration) bool {
	normalizedAddress, err := normalizeAddress(address, port)
	if err != nil {
		logger.Infof("[health check] address format for address \"%s\" failed, %s", address, err.Error())
		return false
	}

	conn, err := net.DialTimeout("tcp", normalizedAddress, timeout)
	if err != nil {
		logger.Infof("[health check] health check for address \"%s\" failed, %s", normalizedAddress, err.Error())
		return false
	}
	defer conn.Close()
	return true
}

// normalizeAddress normalizes the address by ensuring it has the correct port.
// If the port field is empty, it will check if the address already has a port.
//   - If the address has a port, it will return the address as is.
//   - If the address does not have a port, it will return err.
//
// If the port field is not empty, it will check if the address's port matches the provided port.
//   - If it matches, it will return the address as is.
//   - If it does not match, it will return the address with the new port.
func normalizeAddress(address string, port string) (string, error) {
	if port == "" {
		_, _, err := net.SplitHostPort(address)
		if err != nil {
			return "", err
		}
		return address, nil
	}

	host, existingPort, err := net.SplitHostPort(address)
	if err != nil {
		if strings.Contains(err.Error(), "missing port in address") {
			return net.JoinHostPort(strings.Trim(address, "[]"), port), nil
		}
		return "", err
	}

	if existingPort != port {
		return net.JoinHostPort(host, port), nil
	}

	return address, nil
}

```

AlexStocks (approved):

PR #678, opened 2025-05-20
[AI] feat: Add a new field "LLMMeta" for endpoint, simpilify the use of llm endpoint

Alanxtl:
**What this PR does**:

- Add a field `LLMMeta` for endpoint, Allow room for later endpoint extensions specifically for llm.
- Add the mapping from the llm provider to the request url, which increased ease of use for llm users. The step completes the mapping when the endpoint is built.
  In the previous version, if users want to create a llm endpoint, `conf.yaml` should be like this:
  ```yaml
  endpoints:
    - id: 1
      socket_address:
        domains:
          - "api.deepseek.com"
  ```
  After this PR, users can specify the LLM provider. They can do this in `conf.yaml`. The format for this is as follows, which reduces complexity for users:
  ```yaml
  endpoints:
    - id: 1
      llm_meta:
        provider: "deepseek"
  ```
  

After this pr, we can implement the loadbalancer across apiKeys and retry policies with `LLMMeta` field.

Later, more information will be based on the `provider` field, like endpoints, for example:
```yaml
  openai:
    name: OpenAI
    description: OpenAI is an AI research and deployment company. Their mission is to ensure that artificial general intelligence benefits all of humanity.
    base_url: api.openai.com/v1
    endpoints:
      chatComplete: /chat/completions
      imageGenerate: /images/generations
      createSpeech: /audio/speech
```


pkg/model/llmprovider.yaml
```diff
@@ -0,0 +1,38 @@
...
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+
+providers:
+  deepseek:
```
AlexStocks, line 21: 感觉这一行不是很合适，同级别的下面是 openai endpoints 这种广义的名称，这里 deepseek 只是 llm 的一种，你再考虑下

Alanxtl: providers的每一个元素都是一个llm供应商（deepseek、openai、dashscope）
deepseek以下的元素都是对应着这个供应商的metadata比如url和endpoint

AlexStocks: 加一级 llm 吧，在 llm 下面放各个 lm 的信息

Alanxtl: 那是这样的结构吗：
```yaml
providers:
  llm:
    deepseek:
      endpoint:
      url:
    openai:
      endpoint:
      url:
```
这样的话感觉providers和llm是重复的，因为我这里providers指的就是llm供应商，我觉得不需要增加一级llm

AlexStocks (approved):

Issue #645, opened 2024-12-22
Support gRPC steam request


AlexStocks:
related discussion https://github.com/apache/dubbo-go/discussions/2756

AlexStocks:
@zhangcunli

PR #688, opened 2025-06-28
feat(gRPC): Implement Full gRPC Streaming Proxy Functionality with Comprehensive Optimizations

Similarityoung:
ref: #645

**What this PR does**:

1. Core Proxy Functionality

Implemented Passthrough Codec: To address the issue of the proxy's unawareness of message types, a custom grpc.Codec was implemented. It treats all messages as raw bytes ([]byte), effectively bypassing the Protobuf encoding/decoding steps and enabling true "passthrough" proxying.
Unified Stream Handler: Leveraging grpc.UnknownServiceHandler, all gRPC requests, regardless of their type, are directed to a single stream handler. This handler establishes a full-duplex stream to the backend and transparently forwards data between the client and the backend. The request processing path is as follows:
> Client -> GrpcListener -> GrpcProxyConnectionManager -> GrpcProxyFilter -> Backend Service

2. Robustness and Lifecycle Management

Backend Connection Pool & Health Checks: Implemented a thread-safe connection pool (using sync.Map) with locking for backend connections, including a reuse mechanism. A background goroutine was added to monitor connection health and evict stale connections.
Implemented Graceful Shutdown: The framework was extended by adding a Close() method to the GrpcFilter interface. This ensures that all backend connections and resources are safely closed during gateway shutdown (Close/ShutDown), preventing leaks. Idempotency is guaranteed via sync.Once.


Alanxtl:
what's the difference between old `dgp.filter.grpcconnectionmanager` and your `dgp.filter.network.grpcproxy`? It seems the same

Similarityoung:
> what's the difference between old `dgp.filter.grpcconnectionmanager` and your `dgp.filter.network.grpcproxy`? It seems the same

You've raised a very insightful point. From a purely functional standpoint, extending the existing `dgp.filter.grpcconnectionmanager` is certainly feasible.

My decision to create a new `dgp.filter.network.grpcproxy` was primarily driven by the goal of maintaining strict architectural consistency with the existing `dgp.filter.network.dubboproxy` module. As you can see, dubboproxy establishes a clear, network-level proxy pattern that includes a manager and a core filter:

```text
dubboproxy/
├── filter/
│   └── proxy/
│       └── dubbo_proxy_filter.go  (Core proxy logic implementation)
├── dubbofiltermanager.go          (Dubbo filter manager)
├── manager.go                     (Dubbo connection manager & filter chain invocation)
└── plugin.go                      (Plugin entry point for the Dubbo proxy network filter)
```

My `grpcproxy` implementation deliberately mirrors this structure to reduce cognitive load and unify the maintenance patterns for both proxy types.

However, I completely agree with your implicit assessment: placing specific business proxy implementations like `dubboproxy` and `grpcproxy` within the `pkg/filter/network/` directory is not an ideal long-term design.

A clearer refactoring approach that aligns better with the overall project architecture would be to move these reusable proxy capabilities into `pkg/common`. Our generic filter abstractions are already correctly located in `pkg/common/extension/filter`, which shows us the way forward. Therefore, a future evolution path could be:

1. Migrate the `dubboproxy` and `grpcproxy` modules entirely into the `pkg/common/` directory to serve as common proxy implementations.

2. Let `pkg/common/extension/filter` continue to serve as the unified abstraction and management center for all filters.

3. The top-level `pkg/filter/` directory could then be used more purely to orchestrate and apply the different protocol-specific filter chains (like HttpFilter, DubboFilter, GrpcFilter) that are built from the core capabilities in `common`.

This would result in a much cleaner project layering: `pkg/common` provides reusable core capabilities, while `pkg/filter `handles the business-level orchestration. Thank you for your input; it has significantly clarified the path for future refactoring.

Alanxtl:
> > what's the difference between old `dgp.filter.grpcconnectionmanager` and your `dgp.filter.network.grpcproxy`? It seems the same
> 
> You've raised a very insightful point. From a purely functional standpoint, extending the existing `dgp.filter.grpcconnectionmanager` is certainly feasible.
> 
> My decision to create a new `dgp.filter.network.grpcproxy` was primarily driven by the goal of maintaining strict architectural consistency with the existing `dgp.filter.network.dubboproxy` module. As you can see, dubboproxy establishes a clear, network-level proxy pattern that includes a manager and a core filter:
> 
> ```
> dubboproxy/
> ├── filter/
> │   └── proxy/
> │       └── dubbo_proxy_filter.go  (Core proxy logic implementation)
> ├── dubbofiltermanager.go          (Dubbo filter manager)
> ├── manager.go                     (Dubbo connection manager & filter chain invocation)
> └── plugin.go                      (Plugin entry point for the Dubbo proxy network filter)
> ```
> 
> My `grpcproxy` implementation deliberately mirrors this structure to reduce cognitive load and unify the maintenance patterns for both proxy types.
> 
> However, I completely agree with your implicit assessment: placing specific business proxy implementations like `dubboproxy` and `grpcproxy` within the `pkg/filter/network/` directory is not an ideal long-term design.
> 
> A clearer refactoring approach that aligns better with the overall project architecture would be to move these reusable proxy capabilities into `pkg/common`. Our generic filter abstractions are already correctly located in `pkg/common/extension/filter`, which shows us the way forward. Therefore, a future evolution path could be:
> 
> 1. Migrate the `dubboproxy` and `grpcproxy` modules entirely into the `pkg/common/` directory to serve as common proxy implementations.
> 2. Let `pkg/common/extension/filter` continue to serve as the unified abstraction and management center for all filters.
> 3. The top-level `pkg/filter/` directory could then be used more purely to orchestrate and apply the different protocol-specific filter chains (like HttpFilter, DubboFilter, GrpcFilter) that are built from the core capabilities in `common`.
> 
> This would result in a much cleaner project layering: `pkg/common` provides reusable core capabilities, while `pkg/filter `handles the business-level orchestration. Thank you for your input; it has significantly clarified the path for future refactoring.

so you want to move `dgp.filter.network.dubboconnectionmanager` and your `dgp.filter.network.grpcproxy` out of `pkg/filter/` and make them into a proxy level, instead of filtermanager level?

Is my understanding correct? If so, I think the future request processing path would be like this
```mermaid
graph TD
    A[Client] --> B[GrpcListener];
    subgraph p [pixiu]
        B --> C[new proxy level];
        C --> D_sub[old grpcconnectionmanager];
    end

    subgraph D_sub [old grpcconnectionmanager]
        
        D(dgp.filter.grpc.proxy)-->E(other filter)
    end

    E --> G[Backend Service];
```

don't you think the new proxy level and old grpcconnectionmanager have some duplication? and what's the meaning of extract a proxy level are there any functions that cannot be realized now?

Alanxtl:
If I want to test you implemention in the test case in `dubbo-go-pixiu-samples/grpc`
Is my conf.yaml correct?
```yaml
static_resources:
  listeners:
    - name: "grpc"
      protocol_type: "GRPC"
      address:
        socket_address:
          address: "0.0.0.0"
          port: 8881
      filter_chains:
        filters:
          - name: dgp.filter.grpcconnectionmanager
            config:
              route_config:
                routes:
                  - match:
                      prefix: "/provider.UserProvider/"
                    route:
                      cluster: "test-grpc"
                      cluster_not_found_response_code: 505
              http_filters:
                - name: dgp.filter.grpc.proxy
      config:
        idle_timeout: 5s
        read_timeout: 5s
        write_timeout: 5s
  clusters:
    - name: "test-grpc"
      lb_policy: "RoundRobin"
      endpoints:
        - socket_address:
            address: 127.0.0.1
            port: 50001
            protocol_type: "GRPC"
  shutdown_config:
    timeout: "60s"
    step_timeout: "10s"
    reject_policy: "immediacy"
```

Similarityoung:
> If I want to test you implemention in the test case in `dubbo-go-pixiu-samples/grpc` Is my conf.yaml correct?
> 
> ```yaml
> static_resources:
>   listeners:
>     - name: "grpc"
>       protocol_type: "GRPC"
>       address:
>         socket_address:
>           address: "0.0.0.0"
>           port: 8881
>       filter_chains:
>         filters:
>           - name: dgp.filter.grpcconnectionmanager
>             config:
>               route_config:
>                 routes:
>                   - match:
>                       prefix: "/provider.UserProvider/"
>                     route:
>                       cluster: "test-grpc"
>                       cluster_not_found_response_code: 505
>               http_filters:
>                 - name: dgp.filter.grpc.proxy
>       config:
>         idle_timeout: 5s
>         read_timeout: 5s
>         write_timeout: 5s
>   clusters:
>     - name: "test-grpc"
>       lb_policy: "RoundRobin"
>       endpoints:
>         - socket_address:
>             address: 127.0.0.1
>             port: 50001
>             protocol_type: "GRPC"
>   shutdown_config:
>     timeout: "60s"
>     step_timeout: "10s"
>     reject_policy: "immediacy"
> ```

my sample conf is 

```yml
static_resources:
  listeners:
    - name: "grpc"
      protocol_type: "GRPC"
#      protocol_type: "HTTP2"
      address:
        socket_address:
          address: "0.0.0.0"
          port: 8881
      filter_chains:
          filters:
            - name: dgp.filter.network.grpcconnectionmanager
              config:
                route_config:
                  routes:
                    - match:
                        prefix: "/routeguide.RouteGuide/"
                      route:
                        cluster: "test-grpc"
                        cluster_not_found_response_code: 404
                grpc_filters:
                  - name: dgp.filter.grpc.proxy
      config:
        idle_timeout: 5s
        read_timeout: 5s
        write_timeout: 5s
  clusters:
    - name: "test-grpc"
      lb_policy: "RoundRobin"
      endpoints:
        - socket_address:
            address: 127.0.0.1
            port: 50051
            protocol_type: "GRPC"
  shutdown_config:
    timeout: "60s"
    step_timeout: "10s"
    reject_policy: "immediacy"
```

pkg/filter/network/grpcproxy/grpc_manager.go
```diff
@@ -0,0 +1,214 @@
...
+		config:        config,
+		filterManager: filterManager,
+	}
+	gcm.routerCoordinator = router2.CreateRouterCoordinator(&config.RouteConfig)
+	return gcm
+}
+
+// determineStreamType determines the StreamType based on client and server streaming flags
+func determineStreamType(isClientStream, isServerStream bool) grpcCtx.StreamType {
+	if isClientStream && isServerStream {
```
marsevilspirit: use switch?

marsevilspirit:
Give me a test code, I want to verify it.

pkg/context/grpc/context.go
```diff
@@ -0,0 +1,134 @@
...
+	Arguments      []any
+	Attachments    map[string]any
+	Route          *model.RouteAction
+	Result         any
+	Error          error
+	StreamType     StreamType
+	IsStreaming    bool
+	IsStream       bool
+	IsClientStream bool
+	IsServerStream bool
```
marsevilspirit: The fields here are too strange. Why can't we just use StreamType to determine?

marsevilspirit: The fields here are too strange. Why can't we just use StreamType to determine?

pkg/context/grpc/context.go
```diff
@@ -0,0 +1,134 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package grpc
+
+import (
+	"context"
+
+	"github.com/apache/dubbo-go-pixiu/pkg/model"
```
marsevilspirit, line 25: fmt imports.

pkg/common/codec/grpc/passthrough/codec.go
```diff
@@ -0,0 +1,63 @@
...
+	if p, ok := v.(proto.Message); ok {
+		return proto.Unmarshal(data, p)
+	}
+	return fmt.Errorf("passthrough codec: cannot unmarshal into type %T, want *[]byte or proto.Message", v)
+}
+
+// Name returns the name of the codec.
+func (c Codec) Name() string {
+	return "pass_through"
+}
```
marsevilspirit: Why not directly use grpc's codec?

pkg/common/extension/filter/filter.go
```diff
@@ -97,6 +99,15 @@ type (
...
 		OnTripleData(ctx context.Context, methodName string, arguments []any) (any, error)
+		// OnUnaryRPC handles a unary RPC call.
+		OnUnaryRPC(ctx context.Context, fullMethod string, req any) (any, error)
+		// OnStreamRPC handles a streaming RPC call.
+		// In gRPC, a Unary call is fundamentally a special, short-lived case of a stream.
+		// By consolidating all handling here, we leverage a single, robust logic path for all
+		// gRPC proxying, simplifying the architecture and ensuring consistent behavior.
+		OnStreamRPC(stream model.RPCStream, info *model.RPCStreamInfo) error
+		// Close the filter and release resources
+		Close() error
```
marsevilspirit, line 110: I don't think it's appropriate to put these methods in one interface. We can create multiple interfaces and use composition, which is more in line with the Go style.

pkg/filter/network/grpcproxy/filter/proxy/grpc_proxy_filter.go
```diff
@@ -0,0 +1,530 @@
...
+
+	// Start a goroutine to monitor the connection's health.
+	go f.monitorConnection(address, conn)
+
+	return conn, nil
+}
+
+// monitorConnection periodically checks connection health and removes bad connections
+func (f *Filter) monitorConnection(cacheKey string, conn *grpc.ClientConn) {
+	ticker := time.NewTicker(30 * time.Second)
```
marsevilspirit: use constant.

marsevilspirit (commented):
check comment.

pkg/common/codec/grpc/passthrough/codec.go
Alanxtl: What does this method do?
Put "pass_through" in constant and import it, don't hard code.

pkg/model/listener.go
```diff
@@ -71,4 +79,45 @@ type (
...
+	// GrpcConfig gRPC listener specific configuration
+	GrpcConfig struct {
+		MaxReceiveMessageSize int        `yaml:"max_receive_message_size" json:"max_receive_message_size" mapstructure:"max_receive_message_size"`
+		MaxSendMessageSize    int        `yaml:"max_send_message_size" json:"max_send_message_size" mapstructure:"max_send_message_size"`
+		EnableCompression     bool       `yaml:"enable_compression" json:"enable_compression" mapstructure:"enable_compression"`
+		IdleTimeout           string     `yaml:"idle_timeout" json:"idle_timeout" mapstructure:"idle_timeout"`
+		MaxConnectionAge      string     `yaml:"max_connection_age" json:"max_connection_age" mapstructure:"max_connection_age"`
+		EnableTLS             bool       `yaml:"enable_tls" json:"enable_tls" mapstructure:"enable_tls"`
+		TLS                   *TLSConfig `yaml:"tls,omitempty" json:"tls,omitempty" mapstructure:"tls,omitempty"`
+	}
```
Alanxtl: what about set default value in tag

pkg/common/extension/filter/filter.go
```diff
@@ -97,6 +99,15 @@ type (
 		OnData(data any) (any, error)
 		// OnTripleData triple rpc invocation from triple-server
 		OnTripleData(ctx context.Context, methodName string, arguments []any) (any, error)
+		// OnUnaryRPC handles a unary RPC call.
+		OnUnaryRPC(ctx context.Context, fullMethod string, req any) (any, error)
+		// OnStreamRPC handles a streaming RPC call.
+		// In gRPC, a Unary call is fundamentally a special, short-lived case of a stream.
```
Alanxtl, line 105:
```suggestion
		// In gRPC, an Unary call is fundamentally a special, short-lived case of a stream.
```

Similarityoung:
> Give me a test code, I want to verify it.

Sure, please see the test code in this pull request: https://github.com/apache/dubbo-go-pixiu-samples/pull/78

pkg/common/codec/grpc/passthrough/codec.go
Similarityoung: This is a great question. For the current implementation, my goal was to create a transparent pass-through proxy. The logic forwards the data by decoding and then re-encoding it without inspecting or modifying the actual content.

This approach keeps the core logic generic and avoids a direct dependency on specific .proto schema files for now.

However, I agree that using the official gRPC codec is the right way to go when we need to actually manipulate the payload. The plan is to implement that later within a grpcfilter where that kind of processing will be necessary.

pkg/common/extension/filter/filter.go
Similarityoung: Use a instead of 'an' if the following word doesn't start with a vowel sound, e.g. 'a sentence', 'a university'.

pkg/common/codec/grpc/passthrough/codec.go
Similarityoung: to imply codec interface

pkg/filter/network/grpcproxy/filter/proxy/grpc_proxy_filter.go
Similarityoung: done

pkg/context/grpc/context.go
Similarityoung: done

pkg/context/grpc/context.go
Similarityoung: done

Similarityoung: done

pkg/filter/network/grpcproxy/grpc_manager.go
Similarityoung: done

pkg/model/listener.go
Similarityoung: That's a valid point. I've looked through the rest of the codebase and noticed that we don't currently use default tags for setting default values.

pkg/common/extension/filter/filter.go
Alanxtl: my fault😂

marsevilspirit: XD

pkg/common/extension/filter/filter.go
Similarityoung: That's a great point, and I completely agree. Splitting the interface and using composition is definitely more idiomatic Go and better for long-term maintenance.

I took an initial look, and it seems the refactoring for this is quite significant and would go beyond the scope of this current PR.

To address it properly, would it be acceptable to merge this PR first? I can then immediately open a dedicated follow-up PR to refactor the NetworkFilter interface.

Please let me know if this approach works for you.

pkg/model/listener.go
Alanxtl: it depends on whether you need a default value, see pkg/model/base.go:90

Similarityoung: Got it, thank you for the example! I see the established pattern is to use defaults.Set(). I've switched to that method now. Thanks!

AlexStocks:
please fix the file confliction

pkg/common/codec/grpc/passthrough/codec.go
Alanxtl: grpc 有现成的实现

AlexStocks:
the  ci failed

marsevilspirit (approved):
I think this pr could be merged now.

pkg/common/codec/grpc/passthrough/codec.go
```diff
@@ -0,0 +1,66 @@
...
+	if b, ok := v.([]byte); ok {
+		return b, nil
+	}
+	return nil, fmt.Errorf("passthrough codec: cannot marshal type %T, want proto.Message or []byte", v)
+}
+
+// Unmarshal stores the raw data into the target, which must be a *[]byte or proto.Message.
+func (c Codec) Unmarshal(data []byte, v any) error {
+	if vb, ok := v.(*[]byte); ok {
+		*vb = data
```
AlexStocks, line 55: 这行代码有啥意义呢？

Similarityoung: 将接收到的字节流 data存入调用者提供的目标变量中。 

这里的编解码器的设计是它的Marshal 和Unmarshal 都是将字节流不进行任何解析直接传输的。

AlexStocks: 这里 vb 是个临时变量，函数退出后，这个值能保存下来么？我觉得搞个单测测试下

Similarityoung: 这个 vb 就是 v 啊，不过类型是*[]byte，而 v 的类型是 any；*vb = data 这就是把 data 的值放进 v 里面。

AlexStocks (approved):

PR #710, opened 2025-07-28
Fix: fix meritc calc bug

Alanxtl:
the assertion should inside func instead of assert when passing value

### DIFF
--- a/pkg/filter/metric/metric.go
+++ b/pkg/filter/metric/metric.go
@@ -137,7 +137,7 @@ func (f *Filter) Encode(c *http.HttpContext) filter.FilterStatus {
 	return filter.Continue
 }
 
-func computeApproximateResponseSize(res interface{}) (int, error) {
+func computeApproximateResponseSize(res any) (int, error) {
 	if res == nil {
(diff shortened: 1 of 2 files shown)


pkg/filter/metric/metric.go
```diff
@@ -137,11 +137,14 @@ func (f *Filter) Encode(c *http.HttpContext) filter.FilterStatus {
 	return filter.Continue
 }
 
-func computeApproximateResponseSize(res any) (int, error) {
+func computeApproximateResponseSize(res interface{}) (int, error) {
```
Similarityoung: use any？

pkg/prometheus/prometheus.go
```diff
@@ -382,9 +382,12 @@ func computeApproximateRequestSize(r *http.Request) (int, error) {
 	return s, nil
 }
 
-func computeApproximateResponseSize(res *client.UnaryResponse) (int, error) {
+func computeApproximateResponseSize(res interface{}) (int, error) {
```
Similarityoung: use any？

AlexStocks (approved):

PR #876, opened 2026-02-04
Fixes issue #869 where circuit breaker never opens despite backend errors.

### LINKED ISSUE #869: [BUG] HTTP circuit breaker filter fails to trip due to missing error and latency reporting
### ✅ Verification Checklist

- [x] 🔍 I have searched the [existing issues](https://github.com/apache/dubbo-go-pixiu/issues) and confirmed this is not a duplicate

### 🚀 Go Version

1.25.0

### 📦 Dubbo-go-pixiu Version

1.1.0

### 🖥️ Server Configuration

_No response_

### 💻 Client Configuration

_No response_

### 🌐 Protocol Configuration

_No response_

### 📋 Registry Configuration

_No response_

### 💾 Operating System

🪟 Windows

### 📝 Bug Description

The current implementation of` HTTPCircuitBreakerFilter` in pkg/filter/circuitbreaker/circuitbreaker.go is incomplete. While it correctly integrates the Sentinel entry check, it lacks the feedback loop (error and latency reporting) to the Sentinel core.

As a result, the Sentinel circuit breaker state machine will always perceive requests as successful and instantaneous, meaning the circuit breaker will never transition to the OPEN state, regardless of how many 5xx errors or slow responses occur in the actual backend service.

### 🔄 Steps to Reproduce

1.Configure HTTPCircuitBreakerFilter in conf.yaml with a specific resource and a circuit breaker rule (e.g., Strategy: ErrorRatio, Threshold: 0.1).

2.Point the backend service to an endpoint that consistently returns 500 Internal Server Error or has a high latency.

3.Send multiple requests through the Pixiu gateway to this endpoint.

4.Check the Sentinel logs or observe the gateway behavior.

### ✅ Expected Behavior

When the backend service's error ratio or slow call ratio exceeds the configured threshold, Sentinel should transition the resource state to Open. Subsequent requests should be blocked by sentinel.Entry and return a 503 Service Unavailable response locally from the gateway until the circuit breaker enters the Half-Open state.

### ❌ Actual Behavior

The gateway continues to forward all requests to the failing backend service indefinitely. Even if the backend is 100% down, the circuit breaker never trips because` entry.SetError(err) `or `entry.SetSlowRequest(true)` is never called to provide the necessary statistics to the Sentinel state machine.

### 💡 Possible Solution

The filter needs to capture the result of the backend invocation. Since Decode happens before the request is proxied, we should:

Extend the filter to implement the` Encode` (post-processing) phase or use a defer mechanism to capture the final context state.

Check` ctx.BackendError` or the response status code. If it indicates a failure (e.g., 5xx), call `entry.SetError(err)`.

Calculate the latency of the backend call and call `entry.SetSlowRequest(true) `if it exceeds the threshold.

Ensure `entry.Exit() `is called only after these statistics are reported.

LEILEI0628:
**What this PR does**:

This PR fixes the circuit breaker feedback loop implementation in the Sentinel circuit breaker filter. The circuit breaker now correctly transitions to OPEN state when backend services return errors or slow responses.

**Key changes**:
- Modified `Decode()` to store Sentinel entry in HttpContext instead of immediately calling `Exit()`
- Implemented `Encode()` method to handle response processing and error reporting
- Entry-Exit lifecycle now brackets the complete request lifecycle for accurate latency and error statistics
- Circuit breaker now correctly detects error ratios and slow request ratios

Fixes #869

**Root Cause Analysis**:
The original implementation called `entry.Exit()` in the Decode phase (before backend response), causing Sentinel to perceive all requests as successful and instantaneous. The filter lacked error and latency reporting to Sentinel's state machine.

**Testing**:
- Added comprehensive test suite with 7 test cases in `TestCircuitBreakerFeedbackLoop`
- All existing tests continue to pass
- Tested error ratio detection, slow request detection, and circuit breaker state transitions

**Backward Compatibility**:
This change is fully backward compatible. The fix only adds the missing functionality without changing the public API or configuration format.

```release-note
fix(circuitbreaker): Circuit breaker now correctly detects backend errors and slow responses. The circuit breaker will properly transition to OPEN state based on configured error ratio or slow request ratio thresholds, providing effective service protection.
```

### DIFF
--- a/pkg/filter/sentinel/circuitbreaker/circuit_breaker.go
+++ b/pkg/filter/sentinel/circuitbreaker/circuit_breaker.go
@@ -21,20 +21,16 @@ import (
 	"errors"
 	"fmt"
 	"strings"
-)
 
-import (
 	sentinel "github.com/alibaba/sentinel-golang/api"
 	"github.com/alibaba/sentinel-golang/core/circuitbreaker"
-	sc "github.com/alibaba/sentinel-golang/core/config"
-)
 
-import (
 	"github.com/alibaba/sentinel-golang/core/base"
-
+	sc "github.com/alibaba/sentinel-golang/core/config"
 	"github.com/apache/dubbo-go-pixiu/pkg/common/constant"
 	"github.com/apache/dubbo-go-pixiu/pkg/common/extension/filter"
 	"github.com/apache/dubbo-go-pixiu/pkg/context/http"
+
 	pkgs "github.com/apache/dubbo-go-pixiu/pkg/filter/sentinel"
 	"github.com/apache/dubbo-go-pixiu/pkg/logger"
 )
--- a/pkg/filter/sentinel/circuitbreaker/circuit_breaker.go
+++ b/pkg/filter/sentinel/circuitbreaker/circuit_breaker.go
@@ -21,16 +21,19 @@ import (
 	"errors"
 	"fmt"
 	"strings"
+)
 
+import (
 	sentinel "github.com/alibaba/sentinel-golang/api"
-	"github.com/alibaba/sentinel-golang/core/circuitbreaker"
-
 	"github.com/alibaba/sentinel-golang/core/base"
+	"github.com/alibaba/sentinel-golang/core/circuitbreaker"
 	sc "github.com/alibaba/sentinel-golang/core/config"
+)
+
+import (
 	"github.com/apache/dubbo-go-pixiu/pkg/common/constant"
 	"github.com/apache/dubbo-go-pixiu/pkg/common/extension/filter"
 	"github.com/apache/dubbo-go-pixiu/pkg/context/http"
-
 	pkgs "github.com/apache/dubbo-go-pixiu/pkg/filter/sentinel"
 	"github.com/apache/dubbo-go-pixiu/pkg/logger"
 )
--- a/pkg/filter/sentinel/circuitbreaker/circuit_breaker.go
+++ b/pkg/filter/sentinel/circuitbreaker/circuit_breaker.go
@@ -18,6 +18,7 @@
 package circuitbreaker
 
 import (
+	"errors"
 	"fmt"
 	"strings"
 )
@@ -29,6 +30,8 @@ import (
 )
 
 import (
+	"github.com/alibaba/sentinel-golang/core/base"
+
 	"github.com/apache/dubbo-go-pixiu/pkg/common/constant"
 	"github.com/apache/dubbo-go-pixiu/pkg/common/extension/filter"
 	"github.com/apache/dubbo-go-pixiu/pkg/context/http"
@@ -39,6 +42,8 @@ import (
 const (
 	Kind         = constant.HTTPCircuitBreakerFilter
 	Segmentation = "@"
+	// ContextKeySentinelEntry is the key to store Sentinel entry in HttpContext
+	ContextKeySentinelEntry = "sentinel_entry"
 )
 
 func init() {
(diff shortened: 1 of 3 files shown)


Alanxtl:
our project use `import-formatter` to format import blocks, that's the reason why ur CI fails. For you, u should 
1. run `go install github.com/dubbogo/tools/cmd/imports-formatter@latest`
2. cd to the root dir of `dubbo-go-pixiu`
3. run `imports-formatter`

pkg/filter/sentinel/circuitbreaker/circuit_breaker.go
```diff
@@ -101,7 +104,40 @@ func (f *Filter) Decode(ctx *http.HttpContext) filter.FilterStatus {
...
+		ctx.Params = make(map[string]any)
+	}
+	ctx.Params[ContextKeySentinelEntry] = entry
+
+	return filter.Continue
+}
+
+// Encode processes the response and reports statistics to Sentinel
+func (f *Filter) Encode(ctx *http.HttpContext) filter.FilterStatus {
+	entryVal, ok := ctx.Params[ContextKeySentinelEntry]
```
Alanxtl: 这里是不是也要先检查一下
```go
if ctx.Params == nil {
		ctx.Params = make(map[string]any)
	}
```

LEILEI0628: 无需检查，当 ctx.Params == nil 时，ok 为 false，立即 return

pkg/filter/sentinel/circuitbreaker/circuit_breaker.go
```diff
@@ -101,7 +107,40 @@ func (f *Filter) Decode(ctx *http.HttpContext) filter.FilterStatus {
...
+	}
+
+	// Ensure entry.Exit() is called
 	defer entry.Exit()
+
+	// Report error to Sentinel if response indicates failure
+	// Consider 5xx status codes as errors for circuit breaker
+	statusCode := ctx.GetStatusCode()
+	if statusCode >= 500 && statusCode < 600 {
+		entry.SetError(errors.New("backend service error"))
```
Similarityoung: 这里的错误信息是否可以多一点，比如包含实际状态码、错误上下文这种

LEILEI0628: 可以，下次提交一并修改

nanjiek (changes requested):
错误码请参考pkg\context\http里定义

pkg/filter/sentinel/circuitbreaker/circuit_breaker.go
nanjiek: 错误处理机制参考其他filter，具体可以看filter/opa/opa.go

LEILEI0628:
@nanjiek 已更改错误处理、添加更多错误信息、补齐单测

pkg/filter/sentinel/circuitbreaker/circuit_breaker.go
```diff
@@ -39,6 +40,8 @@ import (
 const (
 	Kind         = constant.HTTPCircuitBreakerFilter
 	Segmentation = "@"
+	// ContextKeySentinelEntry is the key to store Sentinel entry in HttpContext
+	ContextKeySentinelEntry = "sentinel_entry"
```
Alanxtl: 把这个移到`pkg/common/constant`里面吧

Alanxtl (approved):
lgtm

nanjiek (approved):
lgtm

PR #919, opened 2026-04-28
fix(cluster): tighten runtime consistency

mochengqian:
### Description
Implements Issue #905 Step 3 only: tighten cluster runtime consistency.
Issue #905 step progress:
- [x] Add tests and benchmarks
- [x] Fix current correctness issues
- [x] Tighten runtime consistency
- [ ] Switch cluster lookup to O(1)
- [ ] Introduce healthy endpoint snapshots
- [ ] Optimize simple LB hot path
- [ ] Optimize consistent-hash LB last

### Why

`ClusterManager` currently keeps both `store.Config` and `clustersMap` runtime state. Before switching cluster lookup to `clustersMap` in a later PR, every write path must keep these views consistent.

Without this step, `UpdateCluster`, `CompareAndSetStore`, and endpoint updates can leave stale runtime clusters, stale healthcheck lifecycle state, or consistent hash entries that no longer match current endpoints.

### Root Cause:

`UpdateCluster` only replaced `store.Config` and did not rebuild/replace the runtime cluster.

`CompareAndSetStore` swapped the store without ensuring the new store had runtime clusters, and without stopping replaced old runtime clusters.

`SetEndpoint` updated endpoint address fields without rebuilding consistent hash, so RingHash/Maglev could still point to old hosts.

Healthcheck runtime stop can happen before checker timers are initialized, causing nil timer panic during early stop.

### PR Goal:

Make cluster write paths maintain a consistent lifecycle boundary between config state and runtime state.

Ensure `store.Config`, `clustersMap`, runtime `Cluster.Config`, healthcheck lifecycle, consistent hash, and round-robin cursor carry-over remain coherent after successful writes.
### Success Criteria:

After `AddCluster` / `UpdateCluster` / `SetEndpoint` / `DeleteEndpoint` / successful `CompareAndSetStore`, `clustersMap[name].Config` points to the current cluster config.

`clustersMap` does not retain stale runtime clusters removed from config.

Replaced runtime clusters are stopped.

`CompareAndSetStore` version mismatch returns `false` with zero side effects.

Round-robin `PrePickEndpointIndex` is atomically carried over during store refresh/update.

Consistent hash reflects current endpoint hosts after endpoint address update/delete.

### Paths this PR should affect:

pkg/server/cluster_manager.go

pkg/server/cluster_manager_test.go

pkg/cluster/healthcheck/healthcheck.go

### Revise

Adding `prepareClusterConfig`: centralizes endpoint assembly and consistent hash creation.

Adding `ensureRuntimeClusters`: repairs nil/missing/mismatched/stale runtime map entries before CAS swap.

Changing `CompareAndSetStore`: ensures runtime clusters, carries over runtime cursor state, swaps store, then stops replaced old runtime clusters.

Changing `UpdateCluster`: rebuilds current config/hash, carries over RR cursor, replaces runtime cluster, and stops the old runtime.

Changing `SetEndpoint` / `DeleteEndpoint`: repairs mismatched runtime state and rebuilds consistent hash from current endpoints.

Adding healthcheck nil timer guard: avoids panic when a runtime cluster is stopped before checker timers are initialized.

Adding regression tests: proves CAS zero side effects, runtime replacement/stop, stale runtime cleanup, cursor carry-over, and direct consistent hash host changes.

### DIFF
--- a/pkg/server/cluster_manager.go
+++ b/pkg/server/cluster_manager.go
@@ -143,16 +143,27 @@ func (cm *ClusterManager) NewStore(version int32) *ClusterStore {
 	return &ClusterStore{Version: version, clustersMap: map[string]*cluster.Cluster{}}
 }
 
+// CompareAndSetStore swaps the store only when versions match.
+// Version mismatch must leave both stores and runtime clusters untouched.
 func (cm *ClusterManager) CompareAndSetStore(store *ClusterStore) bool {
 	cm.rw.Lock()
-	defer cm.rw.Unlock()
 
 	if store.Version != cm.store.Version {
+		cm.rw.Unlock()
 		return false
 	}
 
-	store.carryOverRuntimeStateFrom(cm.store)
+	currentStore := cm.store
+	replacedClusters := store.ensureRuntimeClusters()
+	store.carryOverRuntimeStateFrom(currentStore)
 	cm.store = store
+	if store != currentStore {
+		replacedClusters = append(replacedClusters, currentStore.runtimeClustersNotIn(store)...)
+	}
+	cm.rw.Unlock()
+
+	// Stop old runtime after publishing the swap; Stop may touch timers/goroutines.
+	stopClusters(replacedClusters)
 	return true
 }
 
@@ -249,7 +260,7 @@ func (cm *ClusterManager) RemoveCluster(namesToDel []string) {
 		for _, name := range namesToDel { // suppose resource to remove and clusters is few
 			if name == c.Name {
 				removed := cm.store.Config[i]
-				cm.store.clustersMap[removed.Name].Stop()
+				stopClusters([]*cluster.Cluster{cm.store.clustersMap[removed.Name]})
 				cm.store.Config[i] = nil
 				delete(cm.store.clustersMap, removed.Name)
 			}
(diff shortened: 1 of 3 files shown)


AlexStocks (approved):

pkg/server/cluster_manager.go
```diff
@@ -143,16 +143,27 @@ func (cm *ClusterManager) NewStore(version int32) *ClusterStore {
...
 
-	store.carryOverRuntimeStateFrom(cm.store)
+	currentStore := cm.store
+	replacedClusters := store.ensureRuntimeClusters()
+	store.carryOverRuntimeStateFrom(currentStore)
 	cm.store = store
+	if store != currentStore {
+		replacedClusters = append(replacedClusters, currentStore.runtimeClustersNotIn(store)...)
+	}
+	cm.rw.Unlock()
```
Similarityoung: 建议：这里是否可以把加锁内的逻辑抽成一个小 helper，并在 helper 里用 `defer cm.rw.Unlock()`？

现在手动 `Unlock()` 的逻辑是对的，而且 `stopClusters()` 也保持在锁外执行。但如果后续这里新增其他 return 分支，可能会有漏掉 `Unlock()` 的风险。抽成 helper 后既能保持当前行为，也会更容易维护。

mochengqian: thanks for guidance!

Similarityoung (approved):
LGTM

Alanxtl (approved):
lgtm

PR #930, opened 2026-05-16
fix(endpoint): stabilize endpoint.ID across restarts (nacos + static-config paths)

mochengqian:
❗️Breaking change
### Description
Replaces the random-UUID endpoint identity used by the Nacos LLM registry
and the static cluster-config bootstrap with a deterministic SHA-256-based
scheme, so `endpoint.ID` is stable across process restarts and Nacos
re-subscriptions. Two related bugs in the same Nacos function — `Port`
being overwritten with 0 when `metadata["port"]` fails to parse, and
`Address` not falling back to `instance.Ip` / `instance.Port` — are fixed
along the way.

### Why

`endpoint.ID` is the index key for many runtime and observability paths.
With a random-UUID fallback, every process restart (and every Nacos
re-subscription) produces a fresh ID for the same physical endpoint:

| Path                                                                | What happens after restart                                                          |
| ------------------------------------------------------------------- | ----------------------------------------------------------------------------------- |
| Prometheus metrics with `endpoint` label                            | label values change; each endpoint becomes multiple broken lines in Grafana         |
| Logs correlated by `endpoint.ID`                                    | pre- and post-restart logs no longer link, SRE loses the thread on every restart    |
| `ClusterManager.PickNextEndpoint(curEndpointID)`                    | uses `endpoint.ID` as a retry cursor — every in-flight retry chain drops on restart |
| LLM health-check / circuit-breaker state (indexed by `endpoint.ID`) | treated as a brand-new endpoint, health and breaker state reset to zero             |

This is the kind of bug that doesn't surface in functional tests
(service runs, requests succeed) but bleeds operationally — operators
see endpoints "mysteriously disappearing and reappearing" in dashboards
and end up patching by manually setting `metadata["id"]` on every Nacos
instance. That patches the symptom. This PR removes the cause.

### Root Cause

The relevant `uuid.GenerateUUID()` fallbacks (pre-PR):

```go
// pkg/adapter/llmregistry/registry/nacos/listener.go (before)
if id, ok := instance.Metadata["id"]; ok {
    ret.ID = id
} else {
    ret.ID, _ = uuid.GenerateUUID()   // fresh random UUID every restart
}
```

```go
// pkg/server/cluster_manager.go (before)
if endpoint.ID == "" {
    endpoint.ID, _ = uuid.GenerateUUID()   // same problem in the static-config path
}
```

Neither call site consulted `instance.InstanceId` (Nacos already
provides a stable identifier), and neither used any content-addressable
hash of the endpoint's defining attributes. In the Nacos path the
problem also fires on every re-subscription, not only on restart,
because each callback re-derives `endpoint.ID` from scratch.

Two unrelated bugs sit in the same `generateEndpoint` block:

- `metadata["port"]` parse failure logs a warning but still writes the
  parser's zero result into `Address.Port`, silently zeroing a valid port.
- `Address.Address` / `Address.Port` are populated only from
  `metadata["ip"]` / `metadata["port"]`, so instances that publish only
  `instance.Ip` / `instance.Port` (no metadata overrides) produce
  endpoints with an empty address.

### PR Goal

Make endpoint identity a function of the endpoint's defining attributes
— cluster name, address, provider, API key — instead of a function of
when the process happened to start. Keep behavior consistent between the
Nacos LLM registry path and the static cluster-config bootstrap path so
operators see the same identity rules regardless of registry source.

### Success Criteria

- A Nacos instance carrying `metadata["id"]` continues to use that ID.
- A Nacos instance carrying `InstanceId` but no `metadata["id"]` uses
  `InstanceId` (one upgrade-time shift, then stable forever).
- A Nacos instance carrying neither produces `generated-<sha8>` derived
  from `(metadata["cluster"], address, provider, api_key)`, identical
  across restarts and re-subscriptions.
- A static-config endpoint with empty `id:` produces the same
  deterministic identifier on every boot.
- Renaming an endpoint (`metadata["name"]`) does not change its ID.
- Two endpoints differing only by credential do not collide.
- The raw API key never appears in the generated ID (SHA-256 is one-way).
- `metadata["port"]` parse error preserves the baseline `instance.Port`.
- `Address` baselines from `instance.Ip` / `instance.Port` before
  metadata overrides.

### Upgrade-visible behavior change ⚠

| Source                                | Before           | After              |
| ------------------------------------- | ---------------- | ------------------ |
| Nacos instance with `metadata["id"]`  | `metadata["id"]` | unchanged          |
| Nacos instance with `InstanceId` only | random UUID      | `InstanceId`       |
| Nacos instance without either         | random UUID      | `generated-<sha8>` |
| Static config endpoint with empty ID  | random UUID      | `generated-<sha8>` |

Operators must rebuild any dashboards, log queries, or runbooks keyed on
`endpoint.ID`. On the first deploy after upgrade, downstream consumers
of `OnAddEndpoint` / `OnRemoveEndpoint` will see one round of
remove-then-add per endpoint whose ID changes shape.

There is no runtime opt-out for this behavior change — the migration is
one-shot and reviewers should confirm this is acceptable for their
deployment before merging.

### Known limitations

- **Static endpoint collapse.** Two static endpoints with empty `id:`
  that share the same `(cluster_name, address, provider, api_key)` now
  collapse to a single identifier. They were already indistinguishable
  to the picker; the previous random IDs offered no real separation.
  `assembleClusterEndpoints` logs a `warn` naming both slice positions
  and the shared ID when this happens, so the misconfiguration is
  visible. Set an explicit `id:` to keep them distinct.
- **Nacos instances without `metadata["cluster"]`.** The cluster-name
  component falls back to the empty string, so endpoints across
  different Nacos services sharing `(address, provider, api_key)` will
  collide. Set `metadata.cluster` on the Nacos side to prevent this.

### Paths this PR should affect

- `pkg/model/cluster.go`
- `pkg/model/cluster_test.go`
- `pkg/adapter/llmregistry/registry/nacos/listener.go`
- `pkg/adapter/llmregistry/registry/nacos/listener_test.go`
- `pkg/server/cluster_manager_test.go`
- `go.mod` (drops `hashicorp/go-uuid` from direct dependencies)

### Revise

- **Adding `model.GenerateEndpointID(clusterName, endpoint)`**: returns
  `"generated-<8-byte-hex>"` where the 8 bytes are the leading bytes of
  `sha256(clusterName ‖ address ‖ provider ‖ apiKey)`. Documented design
  notes cover the 64-bit truncation rationale, the empty-clusterName
  edge case, and the deliberate inclusion of the API key.
- **Changing Nacos `generateEndpoint`**: three-tier identity lookup —
  `metadata["id"]` → `instance.InstanceId` → `GenerateEndpointID`. Stops
  zeroing `Address.Port` on `metadata["port"]` parse failure. Seeds
  `Address.Address` / `Address.Port` from `instance.Ip` /
  `instance.Port` before metadata overrides apply.
- **Changing `assembleClusterEndpoints`**: empty `endpoint.ID` is filled
  by `GenerateEndpointID(cluster.Name, endpoint)` instead of
  `uuid.GenerateUUID()`. Updates the comment that incorrectly claimed
  the ID was the endpoint index. Logs a `warn` (with both slice
  positions and the shared ID) whenever two endpoints in the same
  cluster end up with the same generated ID, so the documented collapse
  behavior is loud rather than silent.
- **Hardening `endpointIDMaterial` contract**: the helper now carries a
  comment stating it MUST NOT depend on `endpoint.Name`, because two
  call sites rely on that — `assembleClusterEndpoints` derives the ID
  before assigning a default Name, and the rename-invariance test
  enforces it.
- **Adding regression tests**: deterministic-across-calls, rename
  invariance, credential-difference distinction, address-difference
  distinction, cluster-difference distinction, raw-key non-leakage,
  Nacos `InstanceId` fallback, Nacos generated-hash fallback,
  cross-rename ID stability, static-config endpoint collapse,
  explicit-ID preservation.
- **Removing `hashicorp/go-uuid`** from direct dependencies in `go.mod`;
  it remains transitively pulled in.

### DIFF
--- a/pkg/adapter/llmregistry/registry/nacos/listener.go
+++ b/pkg/adapter/llmregistry/registry/nacos/listener.go
@@ -31,8 +31,6 @@ import (
 
 	"github.com/creasty/defaults"
 
-	"github.com/hashicorp/go-uuid"
-
 	"github.com/nacos-group/nacos-sdk-go/clients/naming_client"
 	nacosModel "github.com/nacos-group/nacos-sdk-go/model"
 	"github.com/nacos-group/nacos-sdk-go/vo"
@@ -275,6 +273,9 @@ func generateEndpoint(instance nacosModel.Instance) *model.Endpoint {
 		return nil
 	}
 
+	ret.Address.Address = instance.Ip
+	ret.Address.Port = int(instance.Port)
+
 	if ip, ok := instance.Metadata["ip"]; ok {
 		ret.Address.Address = ip
 	}
@@ -283,14 +284,9 @@ func generateEndpoint(instance nacosModel.Instance) *model.Endpoint {
 		p, err := strconv.Atoi(port)
 		if err != nil {
 			logger.Warnf("Invalid port in metadata: %s, error: %v", port, err)
+		} else {
+			ret.Address.Port = p
 		}
-		ret.Address.Port = p
-	}
-
-	if id, ok := instance.Metadata["id"]; ok {
-		ret.ID = id
-	} else {
-		ret.ID, _ = uuid.GenerateUUID()
 	}
 
 	if name, ok := instance.Metadata["name"]; ok {
@@ -316,8 +312,19 @@ func generateEndpoint(instance nacosModel.Instance) *model.Endpoint {
 	}
 
 	ret.Metadata = instance.Metadata
+	ret.ID = nacosEndpointID(instance, ret)
 
 	return ret
+}
+
+func nacosEndpointID(instance nacosModel.Instance, endpoint *model.Endpoint) string {
+	if id, ok := instance.Metadata["id"]; ok && strings.TrimSpace(id) != "" {
+		return strings.TrimSpace(id)
+	}
+	if instanceID := strings.TrimSpace(instance.InstanceId); instanceID != "" {
+		return instanceID
+	}
+	return model.GeneratedEndpointID(instance.Metadata["cluster"], endpoint)
 }
 
 func generateInstance(ss nacosModel.SubscribeService) nacosModel.Instance {
--- a/pkg/adapter/llmregistry/registry/nacos/listener.go
+++ b/pkg/adapter/llmregistry/registry/nacos/listener.go
@@ -318,8 +318,8 @@ func generateEndpoint(instance nacosModel.Instance) *model.Endpoint {
 }
 
 func nacosEndpointID(instance nacosModel.Instance, endpoint *model.Endpoint) string {
-	if id, ok := instance.Metadata["id"]; ok && strings.TrimSpace(id) != "" {
-		return strings.TrimSpace(id)
+	if id := strings.TrimSpace(instance.Metadata["id"]); id != "" {
+		return id
 	}
 	if instanceID := strings.TrimSpace(instance.InstanceId); instanceID != "" {
 		return instanceID
--- a/pkg/adapter/llmregistry/registry/nacos/listener.go
+++ b/pkg/adapter/llmregistry/registry/nacos/listener.go
@@ -324,7 +324,11 @@ func nacosEndpointID(instance nacosModel.Instance, endpoint *model.Endpoint) str
 	if instanceID := strings.TrimSpace(instance.InstanceId); instanceID != "" {
 		return instanceID
 	}
-	return model.GenerateEndpointID(instance.Metadata["cluster"], endpoint)
+	clusterName := strings.TrimSpace(instance.Metadata["cluster"])
+	if clusterName == "" {
+		clusterName = strings.TrimSpace(instance.ClusterName)
+	}
+	return model.GenerateEndpointID(clusterName, endpoint)
 }
 
 func generateInstance(ss nacosModel.SubscribeService) nacosModel.Instance {
--- a/pkg/adapter/llmregistry/registry/nacos/listener.go
+++ b/pkg/adapter/llmregistry/registry/nacos/listener.go
@@ -324,7 +324,7 @@ func nacosEndpointID(instance nacosModel.Instance, endpoint *model.Endpoint) str
 	if instanceID := strings.TrimSpace(instance.InstanceId); instanceID != "" {
 		return instanceID
 	}
-	return model.GeneratedEndpointID(instance.Metadata["cluster"], endpoint)
+	return model.GenerateEndpointID(instance.Metadata["cluster"], endpoint)
 }
 
 func generateInstance(ss nacosModel.SubscribeService) nacosModel.Instance {
--- a/pkg/model/cluster.go
+++ b/pkg/model/cluster.go
@@ -130,12 +130,25 @@ func (e Endpoint) GetHost() string {
 	return fmt.Sprintf("%s:%d", e.Address.Address, e.Address.Port)
 }
 
-// GeneratedEndpointID returns a deterministic runtime identity for endpoints
+// GenerateEndpointID returns a deterministic runtime identity for endpoints
 // that do not provide an explicit ID. The hash material includes cluster name,
 // endpoint address, and (when present) LLM provider + API key, so endpoints
 // that differ only by credential do not collide and the same endpoint hashes
 // to the same ID across process restarts.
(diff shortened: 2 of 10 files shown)


pkg/adapter/llmregistry/registry/nacos/listener.go
```diff
@@ -316,10 +312,21 @@ func generateEndpoint(instance nacosModel.Instance) *model.Endpoint {
...
 }
 
+func nacosEndpointID(instance nacosModel.Instance, endpoint *model.Endpoint) string {
+	if id := strings.TrimSpace(instance.Metadata["id"]); id != "" {
+		return id
+	}
+	if instanceID := strings.TrimSpace(instance.InstanceId); instanceID != "" {
+		return instanceID
+	}
+	return model.GenerateEndpointID(instance.Metadata["cluster"], endpoint)
```
AlexStocks: [P1] 这个 tier-3 fallback 只取 `metadata["cluster"]`，但 `generateInstance()` 已经把 Nacos 原生的 `ClusterName` 带进 `instance` 了。若 registry 没把 cluster 再复制到 metadata，同时 `InstanceId` 也为空，不同 cluster、相同 address/credential 的实例会生成同一个 `endpoint.ID`，端点会被错误合并。建议这里至少回退到 `instance.ClusterName`，并补一条“metadata 不含 cluster，但 ClusterName 不同”的回归测试。

mochengqian: ok ,I have resolved this issue. rereview pls.

mochengqian:
@AlexStocks Please merge this bugfix PR when you have time, so I can continue with the rest of the work.Thank you.

AlexStocks (approved):

mochengqian:
Please merge this bugfix PR when you have time, so I can continue with the rest of the work.Thank you.  @Alanxtl

pkg/model/cluster.go
```diff
@@ -128,3 +129,50 @@ func (c *ClusterConfig) CreateConsistentHash() {
...
+//     never alias. Callers from the Nacos LLM path supply
+//     instance.Metadata["cluster"]; if that metadata is missing the value is
+//     the empty string, and identity is then determined by address +
+//     credential only.
+//   - The LLM API key is included on purpose so two endpoints that share an
+//     address but use different credentials never alias to the same ID. The
+//     output is one-way (sha256), so the raw key never appears in the ID.
+func GenerateEndpointID(clusterName string, endpoint *Endpoint) string {
+	sum := sha256.Sum256([]byte(endpointIDMaterial(clusterName, endpoint)))
+	return fmt.Sprintf("generated-%x", sum[:8])
```
Alanxtl:
```suggestion
	return fmt.Sprintf("pixiu-llm-endpoint-%x", sum[:8])
```
前缀加一些语义信息是不是更好

pkg/model/cluster.go
```diff
@@ -128,3 +129,50 @@ func (c *ClusterConfig) CreateConsistentHash() {
...
+	if endpoint == nil {
+		return clusterName
+	}
+	provider := ""
+	apiKey := ""
+	if endpoint.LLMMeta != nil {
+		provider = endpoint.LLMMeta.Provider
+		apiKey = endpoint.LLMMeta.APIKey
+	}
+	return clusterName + "\x00" +
```
Alanxtl: 这里为啥要用\x00作为分隔符

pkg/model/cluster.go
mochengqian: 这个建议我收了，但前缀没有直接写成 `pixiu-llm-endpoint-`，而是改成更通用的 `pixiu-generated-endpoint-`：这个 helper 同时服务静态 cluster 和 LLM registry，两条路径都能读懂，而且日志/面板里一眼就能看出是 Pixiu 自动生成的 ID。

pkg/model/cluster.go
mochengqian: 已采纳这个方向：这里不再依赖裸 `\x00` 分隔，而是改成字段名 + 长度前缀的结构化 material，避免边界歧义，也不影响 hash 的稳定性。

Alanxtl (approved):
lgtm

PR #931, opened 2026-05-16
fix(mcp): clear Content-Length header 

Qiao-yq:
**What this PR does**:

When the MCP filter wraps a backend response into MCP JSON‑RPC format (`sendMCPResponse`, JSON path), the new response body is larger than the original. However, the old `Content-Length` header from the backend response was still copied to the `ResponseWriter`. This caused Go's `net/http` to report `http: wrote more than the declared Content-Length` and forcibly close the connection.

This PR clears the stale `Content-Length` header by calling `ctx.ClearContentLengthHeader()` before returning `filter.Continue` in `sendMCPResponse`. This allows `net/http` to automatically calculate the correct `Content-Length` from the actual response body, consistent with how `sendJSONResponse` and `ErrorHandler.sendResponse` already handle this situation.

Fixes #(https://github.com/apache/dubbo-go-pixiu-samples/issues/142)

- This fix only affects the JSON path (`sendMCPResponse`). The SSE transport path is unchanged and continues to work normally.
- All existing `tools/call` tests pass after this change.
- Verified manually with MCP Inspector: `tools/call` returns a valid JSON‑RPC response.

```release-note
NONE

### DIFF
--- a/pkg/filter/mcp/mcpserver/handlers.go
+++ b/pkg/filter/mcp/mcpserver/handlers.go
@@ -613,6 +613,7 @@ func (f *MCPServerFilter) sendMCPResponse(ctx *MCPContext, response mcp.JSONRPCR
 	ctx.TargetResp = &client.UnaryResponse{Data: mcpResponseBody}
 	ctx.StatusCode(http.StatusOK)
 	ctx.AddHeader(constant.HeaderKeyContextType, constant.HeaderValueApplicationJson)
+	ctx.ClearContentLengthHeader()
 
 	logger.Debugf("[dubbo-go-pixiu] mcp server successfully wrapped backend response in MCP format")
 	return filter.Continue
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: pkg/filter/mcp/mcpserver/handlers.go [hunk 1]
```diff
@@ -613,6 +613,7 @@ func (f *MCPServerFilter) sendMCPResponse(ctx *MCPContext, response mcp.JSONRPCR
 	ctx.TargetResp = &client.UnaryResponse{Data: mcpResponseBody}
 	ctx.StatusCode(http.StatusOK)
 	ctx.AddHeader(constant.HeaderKeyContextType, constant.HeaderValueApplicationJson)
+	ctx.ClearContentLengthHeader()
```
AlexStocks, line 616: [P1] 这个修复点没有配套回归测试。`Content-Length` 不匹配是典型的协议层回归问题，后续重构时这一行很容易再次被删掉。建议在 `mcpserver` 里补一个针对 `sendMCPResponse` 的测试，明确断言包装 backend response 后会清掉 `Content-Length`。

Similarityoung (approved):
LGTM

Alanxtl (approved):
lgtm

Alanxtl:
pls fix ci fail

PR #961, opened 2026-06-05
Feat(cluster): reduce endpoint cloning

### LINKED ISSUE #940: [FEATURE] Avoid double-cloning endpoints during snapshot publication
## Background

PR #932 introduced immutable runtime endpoint snapshots. During cluster config preparation and snapshot publication, endpoints are currently cloned at two different boundaries:

1. `ClusterStore.prepareClusterConfig` / `assembleClusterEndpoints` clones operator-supplied endpoints before assigning default IDs and names.
2. `newEndpointSnapshot` clones config endpoints again so request-path snapshots do not share mutable `Config.Endpoints` objects.

This is correct but can allocate more than necessary when clusters have many endpoints or high membership churn.

## Problem

A single membership update can deep-clone endpoint data twice. That is potentially expensive for large LLM registries or service discovery updates.

However, each clone protects an important boundary:

- Store-side clone prevents Pixiu from mutating caller-owned endpoint objects.
- Snapshot-side clone prevents request-path snapshots from observing later config mutation.

Removing the wrong clone can create subtle compatibility bugs.

## Goal

Reduce avoidable endpoint cloning while preserving the two safety properties above, or prove with benchmark data that the current double clone should remain.

## Suggested approach

1. Add a benchmark that measures membership churn, not only request-path picking. Example shape:
   - 1000 endpoints in one cluster.
   - Repeated `SetEndpoint` calls that update existing endpoint addresses or metadata.
   - Report allocations and ns/op.
2. Compare at least two approaches:
   - Keep store-side clone, make snapshot publication reuse safe already-owned endpoint objects if possible.
   - Keep snapshot-side clone, move ID/name defaulting to a point where caller-owned objects are not mutated.
3. Choose the least surprising behavior, not just the lowest allocation count.
4. Document any externally visible behavior change in CHANGELOG or PR description.

## Files to inspect

- `pkg/server/cluster_manager_bench_test.go`
- `pkg/cluster/cluster.go`
- `pkg/model/cluster.go` (`CloneEndpoint`, `CloneEndpoints`, `GenerateEndpointID`)

## Out of scope

- Do not change endpoint ID generation.
- Do not mutate user-supplied endpoint objects after `AddCluster` or `SetEndpoint`.
- Do not share snapshot-owned endpoint pointers with code that may mutate them.
- Do not rewrite the snapshot mechanism.

## Acceptance criteria

- A benchmark demonstrates the current cost and the improvement, or documents why no safe improvement is worth taking.
- Existing clone-boundary tests continue to pass, especially tests that assert user-supplied endpoints are not mutated.
- Snapshot endpoints remain isolated from later config mutation.
- Tests/benchmarks pass:

```bash
go test ./pkg/server/... ./pkg/cluster/... ./pkg/model/...
go test -bench=. -benchmem ./pkg/server/...
```

## Difficulty

High. This issue is easy to underestimate. It touches compatibility, pointer ownership, and performance. It is better suited for an experienced contributor or for a new contributor paired closely with a maintainer.

ywxzm03:
**What this PR does**:

This PR reduces duplicated endpoint deep clones in the `SetEndpoint` membership churn path, while preserving the existing ownership safety boundaries.

Specifically, this PR:

- Keeps the snapshot-side clone.

- Separates endpoint ID/name defaulting from the endpoint clone boundary.
  - For endpoints already owned by `ClusterStore` in the `SetEndpoint` / `DeleteEndpoint` paths, this PR uses a store-owned prepare path to avoid another full `CloneEndpoints`.
  - It keeps the store-side clone at external input boundaries, so Pixiu still does not mutate endpoint objects provided by callers.

- Adds a membership churn benchmark that measures repeated `SetEndpoint` metadata updates with 1000 endpoints.

This PR strictly follows these constraints:

- Do not change endpoint ID generation.
- Do not mutate user-supplied endpoint objects after `AddCluster` or `SetEndpoint`.
- Do not share snapshot-owned endpoint pointers with code that may mutate them.
- Do not rewrite the snapshot mechanism.

Fixes #940

This PR chooses the second approach suggested in the issue:

```text
Keep snapshot-side clone, move ID/name defaulting to a point where caller-owned objects are not mutated.
```

It does not choose the first approach:

```text
Keep store-side clone, make snapshot publication reuse safe already-owned endpoint objects if possible.
```

Why the first approach was not chosen:

Keeping the store-side clone and allowing snapshots to reuse store-owned endpoints can reduce one deep clone. However, `Cluster.Config` is still a publicly mutable field today. If a snapshot reuses endpoint pointers from `Config.Endpoints`, later direct mutations to `Cluster.Config.Endpoints`, `Metadata`, `LLMMeta`, or `UnHealthy` could pollute an already published runtime snapshot.

To implement the first approach safely, we would likely need changes such as:

- making `Cluster.Config` private and exposing it through read-only or clone-returning getters;
- introducing truly immutable/frozen endpoints;
- splitting the public config from the snapshot-owned endpoint source;
- auditing all config mutation paths and making them copy-on-write.

That scope is much larger and gets close to the “Do not rewrite the snapshot mechanism” boundary. Therefore, this PR does not choose the first approach. Instead, it keeps the snapshot-side clone and only removes the duplicated clone from store-owned mutation paths.

**Verification**:

```bash
go test ./pkg/server/... ./pkg/cluster ./pkg/model/...
go test -run '^$' -bench '^BenchmarkClusterSetEndpointMembershipChurn$' -benchmem ./pkg/server
```

Local benchmark results before and after this change:

Before:

```text
goos: darwin
goarch: arm64
pkg: github.com/apache/dubbo-go-pixiu/pkg/server
cpu: Apple M5
BenchmarkClusterSetEndpointMembershipChurn-10               2100            478004 ns/op         1435526 B/op       9060 allocs/op
PASS
ok      github.com/apache/dubbo-go-pixiu/pkg/server     2.198s
```

After:

```text
goos: darwin
goarch: arm64
pkg: github.com/apache/dubbo-go-pixiu/pkg/server
cpu: Apple M5
BenchmarkClusterSetEndpointMembershipChurn-10               2830            358501 ns/op          947546 B/op       6061 allocs/op
PASS
ok      github.com/apache/dubbo-go-pixiu/pkg/server     1.562s
```

**Does this PR introduce a user-facing change?**

```release-note
NONE
```

### DIFF
--- a/pkg/server/cluster_manager.go
+++ b/pkg/server/cluster_manager.go
@@ -359,9 +359,6 @@ func (s *ClusterStore) AddCluster(c *model.ClusterConfig) {
 // prepareClusterConfig clones operator-supplied endpoints, then rebuilds
 // endpoint defaults and hash from current endpoints.
 func (s *ClusterStore) prepareClusterConfig(c *model.ClusterConfig) {
-	if c == nil {
-		return
-	}
 	c.Endpoints = model.CloneEndpoints(c.Endpoints)
 	s.prepareOwnedClusterConfig(c)
 }
@@ -370,20 +367,17 @@ func (s *ClusterStore) prepareClusterConfig(c *model.ClusterConfig) {
 // already owned by ClusterStore. Callers must not pass operator-owned endpoint
 // pointers here; use prepareClusterConfig at external input boundaries.
 func (s *ClusterStore) prepareOwnedClusterConfig(c *model.ClusterConfig) {
-	if c == nil {
-		return
-	}
 	s.assembleClusterEndpoints(c)
 	c.CreateConsistentHash()
 }
 
-// assembleClusterEndpoints assembles the cluster endpoints by formatting the
-// ID, name and domains for each endpoint. If endpoint.LLMMeta is not nil, the
-// assimilation of name and domain is based on the LLM provider denoted in the
-// endpoint LLMMeta. Callers choose the ownership boundary before invoking this
-// helper: external input paths clone endpoints in prepareClusterConfig, while
-// store-owned mutation paths call prepareOwnedClusterConfig to avoid a second
-// full endpoint clone before snapshot publication.
+// assembleClusterEndpoints assembles the cluster endpoints by assigning stable
+// unique IDs and default names for each endpoint. If endpoint.LLMMeta is not nil,
+// the default endpoint name is based on the LLM provider denoted in the endpoint
+// LLMMeta. Callers choose the ownership boundary before invoking this helper:
+// external input paths clone endpoints in prepareClusterConfig, while store-owned
+// mutation paths call prepareOwnedClusterConfig to avoid a second full endpoint
+// clone before snapshot publication.
 func (s *ClusterStore) assembleClusterEndpoints(c *model.ClusterConfig) {
 	if c == nil {
 		return
--- a/pkg/server/cluster_manager.go
+++ b/pkg/server/cluster_manager.go
@@ -356,8 +356,23 @@ func (s *ClusterStore) AddCluster(c *model.ClusterConfig) {
 	stopClusters([]*cluster.Cluster{s.replaceClusterRuntime(c.Name, c)})
 }
 
-// prepareClusterConfig rebuilds endpoint defaults and hash from current endpoints.
+// prepareClusterConfig clones operator-supplied endpoints, then rebuilds
+// endpoint defaults and hash from current endpoints.
 func (s *ClusterStore) prepareClusterConfig(c *model.ClusterConfig) {
+	if c == nil {
+		return
+	}
+	c.Endpoints = model.CloneEndpoints(c.Endpoints)
+	s.prepareOwnedClusterConfig(c)
+}
+
+// prepareOwnedClusterConfig rebuilds endpoint defaults and hash for endpoints
+// already owned by ClusterStore. Callers must not pass operator-owned endpoint
+// pointers here; use prepareClusterConfig at external input boundaries.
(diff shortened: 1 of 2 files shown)


pkg/server/cluster_manager.go
```diff
@@ -942,7 +952,7 @@ func (s *ClusterStore) DeleteEndpoint(clusterName string, endpointID string) {
 		if e.ID == endpointID {
 			runtimeCluster.RemoveEndpoint(e)
 			clusterConfig.Endpoints = append(clusterConfig.Endpoints[:i], clusterConfig.Endpoints[i+1:]...)
-			s.prepareClusterConfig(clusterConfig)
+			s.prepareOwnedClusterConfig(clusterConfig)
```
mochengqian: DeleteEndpoint 现在删除后直接走 prepareOwnedClusterConfig，不再像旧路径那样通过 prepareClusterConfig 全量 clone。这样 append(clusterConfig.Endpoints[:i], clusterConfig.Endpoints[i+1:]...) 缩短 slice 后不会清理 backing array 的尾部指针。删除尾部 endpoint，尤其是连续删除尾部时，已删除 endpoint 及其 Metadata / LLMMeta 可能被 backing array 继续持有，直到 cluster 被替换或 slice 重新分配。

ywxzm03: 引入了slices.Delete来解决删除中间元素和结尾元素带来的内存泄露隐患，但需要最小go版本1.21，本项目版本为1.25.0；如果需要兼容老版本，也可以改为手写的slices.Delete

AlexStocks (approved):

PR #932, opened 2026-05-17
feat(cluster):introduce healthy endpoint snapshot 

mochengqian:
## Summary

Migrates the cluster runtime endpoint view from mutable `Config.Endpoints` +
`Endpoint.UnHealthy` into an immutable `EndpointSnapshot` published via
`atomic.Pointer`. Request-path picks read snapshots; health updates and
membership changes go through CAS replacement instead of mutating shared
state.
## Architecture

- `Cluster.endpoints atomic.Pointer[EndpointSnapshot]` publishes the
  immutable view. Health updates produce a new snapshot via CAS.
- `EndpointSnapshot` indexes endpoints by ID, address, and health, all
  precomputed at publish time so the request path is O(1) on the healthy
  view instead of an O(N) scan over `ClusterConfig.Endpoints`.
- `PickContext` carries snapshot + config into load balancers. Bundled
  balancers (round-robin, rand, weight-random, ring-hash, maglev)
  implement `SnapshotLoadBalancer`; legacy balancers keep working via
  the compatibility adapter, which serializes them through a
  package-level lock.
- `healthcheck.HealthChecker` emits `EndpointHealthEvent` via callback;
  the legacy direct-mutation path (`CreateHealthCheck` without a
  callback) is preserved for existing callers.
- LLM proxy cooldown state moved out of `Endpoint.Metadata` into a
  process-wide bounded store (`maxCooldownStoreEntries = 1024`) with a
  throttled sweep.

## Operator-visible behavior changes

(For the maintainer writing release notes — these belong in the
next CHANGELOG entry.)

- **Endpoint state is immutable post-publish.** Direct mutation of
  `Endpoint.UnHealthy` or `Endpoint.Metadata` after cluster start is no
  longer observed by the request path. Health updates go through
  `Cluster.UpdateEndpointHealth` / health-check callbacks; membership
  changes go through `Cluster.RefreshEndpoints`.
- **`Endpoint.UnHealthy` post-init mutation is no longer observed by the
  request path.** Code that flips `endpoint.UnHealthy = true/false` after
  the cluster has started will see no effect on which endpoints get picked;
  the snapshot is built once and only updated via
  `Cluster.UpdateEndpointHealth` (or via the health-check callback).
  Upgrade impact: any external integration that flipped this field
  directly will silently stop functioning. The healthcheck legacy path
  (`CreateHealthCheck` without callback) still mutates the field for
  backward compatibility, but pickers do not read it.
[...]

PR #962, opened 2026-06-07
feat(cluster): add metrics for endpoint snapshot publication

### LINKED ISSUE #944: [FEATURE] Add metrics for cluster endpoint snapshot publication
## Background

PR #932 introduced immutable endpoint snapshots for cluster runtime state. Operators currently do not have direct metrics for snapshot publish frequency or snapshot endpoint size.

Without these metrics, registry churn, unexpected cluster size, or excessive health update publication can be hard to diagnose.

## Goal

Add observability for endpoint snapshot publication.

Suggested metrics:

- `pixiu_cluster_snapshot_publish_total`: counter incremented when a new snapshot is successfully published.
- `pixiu_cluster_snapshot_endpoint_count`: gauge for total endpoints in the latest published snapshot.
- Optional if it matches existing project conventions: healthy endpoint count gauge.

Metric names and labels should follow existing Pixiu metric style.

## Suggested approach

1. Inspect existing metrics definitions and registration patterns in the repository.
2. Add the smallest metrics surface that follows current naming and label conventions.
3. Emit metrics from the snapshot publication path, likely around successful `CompareAndSwap` in cluster snapshot refresh/update methods.
4. Add tests if the project already has metric tests; otherwise add a focused test around metric collection if practical.
5. Update docs only if there is an existing metrics reference page for similar cluster metrics.

## Files to inspect

- `pkg/cluster/cluster.go`
- Existing metrics packages/usages under `pkg/`
- Existing observability or prometheus tests, if present.

## Out of scope

- Do not add a new metrics backend.
- Do not introduce high-cardinality labels such as endpoint ID or endpoint address.
- Do not log every snapshot publish as a substitute for metrics.
- Do not change snapshot publication behavior.

## Acceptance criteria

- Metrics are registered using the repository's existing metrics style.
- Snapshot publish counter increments only on successful publish.
- Endpoint count reflects the latest published snapshot size.
- Labels avoid endpoint-level cardinality.
- Tests pass:

```bash
go test ./pkg/cluster/...
go test ./pkg/server/...
go test ./...
```

## Difficulty

Medium. This is a good community issue if the contributor first follows existing metrics patterns. The main review focus will be label cardinality and avoiding behavior changes.

mochengqian:
## What

Adds observability for endpoint snapshot publication (follow-up to #932). Operators currently have no direct signal for snapshot publish frequency or snapshot endpoint size, which makes registry churn, unexpected cluster size, or excessive health-update publication hard to diagnose.

## Metrics

Three OpenTelemetry instruments, emitted on each successful snapshot `CompareAndSwap`, labeled only by `cluster`:

| Metric | Type | Meaning |
|--------|------|---------|
| `pixiu_cluster_snapshot_publish_total` | counter | Incremented once per successful snapshot publish |
| `pixiu_cluster_snapshot_endpoint_count` | gauge | Total endpoints in the latest published snapshot |
| `pixiu_cluster_snapshot_healthy_endpoint_count` | gauge | Healthy endpoints in the latest published snapshot |

Instruments follow the existing `pixiu_`/snake_case convention and bind lazily to `otel.GetMeterProvider()` (same pattern as `pkg/filter/metric` and the LLM tokenizer), so they wire into the existing Prometheus exporter automatically.

## Where it emits

`recordSnapshotPublish` is called inside the successful-CAS branch of the three publish paths in `pkg/cluster/cluster.go`:

- `RefreshEndpointsFrom` — membership/address changes
- `UpdateEndpointHealth` — endpoint-keyed health flips
- `UpdateEndpointAddressHealth` — address-keyed health flips

The counter increments **only** on a real swap: no-op health updates return before CAS and are not counted, which is what makes "excessive health update publication" diagnosable.

## Design notes / trade-offs

- **Synchronous emission, not an `ObservableGauge`.** A callback gauge would require a global live-cluster registry to enumerate clusters at scrape time; this PR avoids introducing new global mutable state and emits at the publish site instead.
- **Stale series on cluster deletion.** Because emission is synchronous, a deleted cluster's gauge retains its last value until the process restarts. Clusters are long-lived configuration objects with rare name churn, so this is an accepted trade-off rather than a leak.
- **One attribute set allocated per publish** (`WithAttributes`). Publish is a low-frequency path (not the pick hot path), so this is not cached; the read/pick path is untouched.
- **Scope held deliberately tight:** no duration histogram, generation gauge, or per-state breakdown — those can follow as separate issues.

## Out of scope (per #944)

No new metrics backend, no endpoint-level (high-cardinality) labels, and no change to snapshot publication behavior.

## Tests

`pkg/cluster/metrics_test.go` uses a `ManualReader` to assert:

- publish counter and endpoint/healthy gauges after an endpoint-health flip
- no-op health update does **not** increment the counter
- address-keyed health flip publishes once and marks all endpoints sharing the address

`pkg/server/cluster_manager_test.go` includes `TestStaticClusterSnapshotMetricsRecordedAtStartup` to guard against startup ordering regression (see fix below).

```bash
go test ./pkg/cluster/... ./pkg/server/...
go vet ./pkg/cluster/
```

All pass; `gofmt` clean.

---

## 🔧 Fix: Startup ordering for static clusters

**Problem found during review:** Static clusters' initial snapshot publish was landing on the no-op delegating provider, causing counter/gauges to remain empty in steady state — exactly the "unexpected cluster size" scenario this PR is meant to diagnose.

**Root cause:** `Start(bs)` called `server.initialize(bs)` (which constructs clusters) before `registerOtelMetricMeter()`. While otel-go's delegation back-fills instruments when `SetMeterProvider` is called later, the data from those dropped emissions is lost forever.

**Fix applied (commit `7ac917f`):**
- Moved `registerOtelMetricMeter(bs.Metric)` from `(*Server).Start()` to `Start(bs)`, positioned before `server.initialize(bs)`
- Added `TestStaticClusterSnapshotMetricsRecordedAtStartup` in `pkg/server` to guard against ordering regression
- Expanded test helper documentation about global mutation constraints

**Verification:** All tests pass including the new ordering guard; local CI clean (gofmt, imports, golangci-lint, race detector).

Closes #944

### DIFF
--- a/pkg/cluster/metrics.go
+++ b/pkg/cluster/metrics.go
@@ -1,0 +1,87 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package cluster
+
+import (
+	"context"
+	"sync"
+)
+
+import (
+	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/metric"
+)
+
+import (
+	"github.com/apache/dubbo-go-pixiu/pkg/logger"
+)
+
+// Endpoint snapshot publication metrics. Instruments are bound lazily to the
+// global MeterProvider on first publish; when metrics are disabled the provider
+// is a no-op, so recording stays a cheap no-op and never blocks publication.
+var (
+	snapshotMetricsOnce sync.Once
+
+	snapshotPublishTotal  metric.Int64Counter
+	snapshotEndpointCount metric.Int64Gauge
+	snapshotHealthyCount  metric.Int64Gauge
+)
+
+func initSnapshotMetrics() {
+	snapshotMetricsOnce.Do(func() {
+		meter := otel.GetMeterProvider().Meter("pixiu")
+
+		var err error
+		if snapshotPublishTotal, err = meter.Int64Counter("pixiu_cluster_snapshot_publish_total",
+			metric.WithDescription("Total number of cluster endpoint snapshots successfully published.")); err != nil {
+			logger.Errorf("[dubbo-go-pixiu] register pixiu_cluster_snapshot_publish_total failed: %v", err)
+		}
+		if snapshotEndpointCount, err = meter.Int64Gauge("pixiu_cluster_snapshot_endpoint_count",
+			metric.WithDescription("Total endpoints in the latest published cluster snapshot.")); err != nil {
+			logger.Errorf("[dubbo-go-pixiu] register pixiu_cluster_snapshot_endpoint_count failed: %v", err)
+		}
+		if snapshotHealthyCount, err = meter.Int64Gauge("pixiu_cluster_snapshot_healthy_endpoint_count",
+			metric.WithDescription("Healthy endpoints in the latest published cluster snapshot.")); err != nil {
+			logger.Errorf("[dubbo-go-pixiu] register pixiu_cluster_snapshot_healthy_endpoint_count failed: %v", err)
+		}
+	})
+}
+
+// recordSnapshotPublish reports a successful snapshot publication for a cluster.
+// Call exactly once per successful CompareAndSwap of the published snapshot. The
+// only label is the cluster name, which is bounded by configuration rather than
+// request traffic, so cardinality stays low. The health-update callers invoke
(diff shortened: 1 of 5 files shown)


AlexStocks (approved):

pkg/cluster/metrics.go
```diff
@@ -0,0 +1,87 @@
...
+// allocation-light and must not block.
+func recordSnapshotPublish(clusterName string, snapshot *EndpointSnapshot) {
+	initSnapshotMetrics()
+
+	attrs := metric.WithAttributes(attribute.String("cluster", clusterName))
+	ctx := context.Background()
+
+	if snapshotPublishTotal != nil {
+		snapshotPublishTotal.Add(ctx, 1, attrs)
+	}
```
AlexStocks, line 80: [P1] initSnapshotMetrics 第一次调用会在 healthMu 持有期间执行：UpdateEndpointHealth/UpdateEndpointAddressHealth 都在 healthMu.Lock() 之后直接走到 recordSnapshotPublish → initSnapshotMetrics → meter.Int64Counter/Int64Gauge × 3，注册失败时还要走 logger.Errorf，整个路径的分配和潜在 IO 全部发生在锁内。这与本函数注释“health-update callers invoke this while holding the cluster s healthMu, so the body must stay allocation-light and must not block”直接冲突，会让首个健康事件、或者每次 registerOtelMeterProvider 之后第一次 publish 抢占 healthMu，阻塞所有并发健康事件。建议在 cluster 包初始化（或 registerOtelMetricMeter 完成后）显式调用一次 initSnapshotMetrics 把仪表预绑定好，让 recordSnapshotPublish 只剩 Add/Record。

Similarityoung (approved):
lgtm

PR #965, opened 2026-06-07
perf(llm): drop per-request hex alloc in cooldown credential hash

### LINKED ISSUE #956: [FEATURE] Avoid per-request hex allocation in LLM cooldown credential hash
## Background

Follow-up to #932 (issue #905 step 5), which moved LLM proxy cooldown state out of `Endpoint.Metadata` into a process-wide `cooldownStore` keyed by `cooldownKey`.

In `pkg/filter/llm/proxy/filter.go`:

```go
cooldownKey struct {
	clusterName     string
	endpointID      string
	endpointAddress string
	credentialHash  string
}

func endpointCredentialHash(endpoint *model.Endpoint) string {
	if endpoint == nil || endpoint.LLMMeta == nil || endpoint.LLMMeta.APIKey == "" {
		return ""
	}
	sum := sha256.Sum256([]byte(endpoint.LLMMeta.APIKey))
	return fmt.Sprintf("%x", sum)
}
```

## Problem

Every LLM request flows through `endpointInCooldown` / `markEndpointCooldown` -> `newCooldownKey` -> `endpointCredentialHash` at least once (more on the retry/fallback path). Each call:

- recomputes sha256 of the API key, and
- `fmt.Sprintf("%x", sum)` allocates a fresh 64-byte hex string that escapes to the heap.

The hex string is pure overhead — it is only used as part of a comparable map key.

## Goal

Remove the per-request hex-string allocation.

## Suggested approach

Use the raw `[32]byte` sha256 output directly as the key component (it is a comparable array and is valid in a struct map key):

```go
cooldownKey struct {
	clusterName     string
	endpointID      string
	endpointAddress string
	credentialHash  [32]byte
}

func endpointCredentialHash(endpoint *model.Endpoint) [32]byte {
	if endpoint == nil || endpoint.LLMMeta == nil || endpoint.LLMMeta.APIKey == "" {
		return [32]byte{}
	}
	return sha256.Sum256([]byte(endpoint.LLMMeta.APIKey))
}
```

This drops the hex encoding + heap allocation. sha256 itself stays per-call (nanoseconds for short keys); the allocation was the dominant cost. Caching the hash on the endpoint is explicitly **not** proposed — it would pollute the model struct and complicate snapshot cloning for a ~200ns win.

## Test plan

- [ ] `BenchmarkCooldown_EndpointInCooldown` over ~100 LLM endpoints, before/after — expect at least one fewer alloc/op.
- [ ] Update existing cooldown tests for the `[32]byte` zero value.
- [ ] `go test -race ./pkg/filter/llm/proxy/...`

## Files

- `pkg/filter/llm/proxy/filter.go`
- `pkg/filter/llm/proxy/filter_test.go`

## Scope note

`cooldownKey` is unexported; this is a purely internal change with no upstream-visible API impact. Independent of #939 (inject cooldownStore) and #943 (container/list LRU) — can land in any order.

mochengqian:
## Summary

Closes #956.

Every LLM request resolves a `cooldownKey` at least once (more on the retry/fallback path) through `endpointCredentialHash`. That function ran `fmt.Sprintf("%x", sum)` and allocated a fresh 64-byte hex string that escapes to the heap. The hex string is pure overhead — `credentialHash` is only ever used as a comparable component of the `cooldownKey` map key, never displayed or logged.

This PR uses the raw `[32]byte` sha256 output directly as the key component. A `[32]byte` array is comparable and valid in a struct map key, so the hex encoding and its heap allocation are removed while the cooldown identity semantics stay byte-for-byte identical.

`sha256` itself stays per-call (nanoseconds for short keys); the allocation was the dominant cost. Caching the hash on the endpoint was explicitly **not** done — it would pollute the model struct and complicate snapshot cloning for a ~200ns win, as the issue notes.

## Change

- `cooldownKey.credentialHash`: `string` → `[32]byte`
- `endpointCredentialHash(...)`: returns `[32]byte` (zero value for no-credential endpoints), dropping `fmt.Sprintf`

`cooldownKey` is unexported, so this is a purely internal change with no upstream-visible API impact.

## Benchmark

`BenchmarkCooldown_EndpointInCooldown` (added in this PR), 100 LLM endpoints each with an API key, on the hot `endpointInCooldown` path:

| | ns/op | B/op | allocs/op |
|---|---|---|---|
| before | 300.0 | 168 | 6 |
| after | 163.2 | 40 | 3 |

3 fewer allocs/op and ~46% faster, exceeding the issue's "at least one fewer alloc/op" target.

## Test plan

- [x] `go test -race ./pkg/filter/llm/proxy/...` — existing cooldown tests pass unchanged with the `[32]byte` zero value
- [x] `go vet ./pkg/filter/llm/proxy/` clean
- [x] `go build ./...` clean
- [x] New `BenchmarkCooldown_EndpointInCooldown` documents the before/after alloc delta

### DIFF
--- a/pkg/filter/llm/proxy/filter.go
+++ b/pkg/filter/llm/proxy/filter.go
@@ -118,7 +118,7 @@ type (
 		clusterName     string
 		endpointID      string
 		endpointAddress string
-		credentialHash  string
+		credentialHash  [32]byte
 	}
 
 	cooldownStore struct {
@@ -582,12 +582,11 @@ func newCooldownKey(clusterName string, endpoint *model.Endpoint) cooldownKey {
 	}
 }
 
-func endpointCredentialHash(endpoint *model.Endpoint) string {
+func endpointCredentialHash(endpoint *model.Endpoint) [32]byte {
 	if endpoint == nil || endpoint.LLMMeta == nil || endpoint.LLMMeta.APIKey == "" {
-		return ""
+		return [32]byte{}
 	}
-	sum := sha256.Sum256([]byte(endpoint.LLMMeta.APIKey))
-	return fmt.Sprintf("%x", sum)
+	return sha256.Sum256([]byte(endpoint.LLMMeta.APIKey))
 }
 
 // endpointCooldownInterval returns the per-endpoint cooldown TTL derived from
--- a/pkg/filter/llm/proxy/filter_test.go
+++ b/pkg/filter/llm/proxy/filter_test.go
@@ -412,3 +412,26 @@ func testLLMEndpoint(id string, port int) *model.Endpoint {
 		},
 	}
 }
+
+func BenchmarkCooldown_EndpointInCooldown(b *testing.B) {
+	const clusterName = "llm-cooldown-bench"
+	const endpointCount = 100
+	store := newCooldownStore()
+	executor := &RequestExecutor{
+		clusterName: clusterName,
+		cooldowns:   store,
+	}
+	endpoints := make([]*model.Endpoint, endpointCount)
+	for i := range endpoints {
+		endpoint := testLLMEndpoint(fmt.Sprintf("ep-%d", i), 19000+i)
+		endpoint.LLMMeta.APIKey = fmt.Sprintf("api-key-%d", i)
+		endpoints[i] = endpoint
+		store.markFailure(clusterName, endpoint, time.Now())
+	}
+
+	b.ReportAllocs()
+	b.ResetTimer()
+	for i := 0; i < b.N; i++ {
+		executor.endpointInCooldown(endpoints[i%endpointCount])
+	}
+}
--- a/pkg/filter/llm/proxy/filter_test.go
+++ b/pkg/filter/llm/proxy/filter_test.go
@@ -416,6 +416,10 @@ func testLLMEndpoint(id string, port int) *model.Endpoint {
 func BenchmarkCooldown_EndpointInCooldown(b *testing.B) {
 	const clusterName = "llm-cooldown-bench"
 	const endpointCount = 100
+	// Keep the cooldown TTL far longer than any -benchtime run so every
+	// iteration stays on the intended in-cooldown hot path. Otherwise entries
+	// could expire mid-benchmark and shift measurement onto the delete+log path.
+	const cooldownTTLMillis = int64(24 * time.Hour / time.Millisecond)
 	store := newCooldownStore()
 	executor := &RequestExecutor{
 		clusterName: clusterName,
@@ -425,6 +429,7 @@ func BenchmarkCooldown_EndpointInCooldown(b *testing.B) {
 	for i := range endpoints {
 		endpoint := testLLMEndpoint(fmt.Sprintf("ep-%d", i), 19000+i)
 		endpoint.LLMMeta.APIKey = fmt.Sprintf("api-key-%d", i)
+		endpoint.LLMMeta.HealthCheckInterval = cooldownTTLMillis
 		endpoints[i] = endpoint
 		store.markFailure(clusterName, endpoint, time.Now())
 	}


pkg/filter/llm/proxy/filter_test.go
```diff
@@ -412,3 +412,26 @@ func testLLMEndpoint(id string, port int) *model.Endpoint {
...
+	store := newCooldownStore()
+	executor := &RequestExecutor{
+		clusterName: clusterName,
+		cooldowns:   store,
+	}
+	endpoints := make([]*model.Endpoint, endpointCount)
+	for i := range endpoints {
+		endpoint := testLLMEndpoint(fmt.Sprintf("ep-%d", i), 19000+i)
+		endpoint.LLMMeta.APIKey = fmt.Sprintf("api-key-%d", i)
+		endpoints[i] = endpoint
```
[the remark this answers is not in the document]
mochengqian: Good catch — fixed in f2aa49be. The benchmark now sets each endpoint's `HealthCheckInterval` to 24h (via a `cooldownTTLMillis` const), so the cooldown TTL far exceeds any realistic `-benchtime`. Every iteration stays on the intended in-cooldown hot path and can't fall through to the delete+log path mid-run. Verified stable at `-benchtime=2s` (~13.5M iters): 164.8 ns/op, 40 B/op, 3 allocs/op.

pkg/filter/llm/proxy/filter.go
```diff
@@ -582,12 +582,11 @@ func newCooldownKey(clusterName string, endpoint *model.Endpoint) cooldownKey {
 	}
 }
 
-func endpointCredentialHash(endpoint *model.Endpoint) string {
```
AlexStocks, line 585: 没看明白这里改动的意义在哪

mochengqian: 意义是减少热路径 allocation。
原589: sum := sha256.Sum256([]byte(endpoint.LLMMeta.APIKey))
原590: return fmt.Sprintf("%x", sum)
这里64 字节 hex string，通常会分配,要是换成[32]byte，可直接作为map key.
原来的 hex string 只是为了让 hash 可比较，但 Go 的 [32]byte 本来就可比较，也能放进 struct map key。所以 hex 编码这一步不就是纯开销了吗。删掉它可以少一次格式化和 heap allocation。
您要是觉得没意义我就把PR关了 @AlexStocks

AlexStocks: 最好给一个 benchmark 测试结果

mochengqian:
已有benchmark:
`BenchmarkCooldown_EndpointInCooldown` (added in this PR), 100 LLM endpoints each with an API key, on the hot `endpointInCooldown` path:

| | ns/op | B/op | allocs/op |
|---|---|---|---|
| before | 300.0 | 168 | 6 |
| after | 163.2 | 40 | 3 |

补充benchmark @AlexStocks :
验证长时间稳定性 (`-benchtime=10s`)输出:
BenchmarkCooldown_EndpointInCooldown-10   71128392   166.6 ns/op   40 B/op   3 allocs/op
结果:71M+迭代运行稳定

验证多核并发 (`-cpu=1,2,4,8`)输出:
BenchmarkCooldown_EndpointInCooldown     6498583   187.3 ns/op   40 B/op   3 allocs/op
BenchmarkCooldown_EndpointInCooldown-2   7174485   167.5 ns/op   40 B/op   3 allocs/op
BenchmarkCooldown_EndpointInCooldown-4   7150099   167.9 ns/op   40 B/op   3 allocs/op
BenchmarkCooldown_EndpointInCooldown-8   7156354   167.5 ns/op   40 B/op   3 allocs/op
结果:2-8核心表现一致

验证内存:
命令: $ go tool pprof -top -alloc_objects mem.out ,输出:
 
 12026039 62.36%  SocketAddress.GetAddress
   7143532 37.04%  fmt.Sprintf  

结果:本PR消除的hex编码部分的fmt.Sprintf得到的优化,已验证

AlexStocks (approved):

PR #966, opened 2026-06-07
perf(llm): make cooldown store eviction O(1) with container/list LRU

### LINKED ISSUE #943: [FEATURE] Replace LLM cooldown eviction scan with container/list LRU
## Background

The LLM proxy cooldown store is bounded by `maxCooldownStoreEntries`. When it reaches capacity, it currently evicts the oldest entry by scanning the map.

This is simple and correct for a small cap, but eviction is O(N) on the failure path.

## Problem

The cooldown store should keep predictable behavior as the number of tracked endpoints grows. An explicit LRU list would make eviction O(1) while preserving the current key and TTL behavior.

## Goal

Replace scan-based oldest-entry eviction with a small LRU structure based on Go's standard `container/list` package.

## Suggested approach

1. Extend `cooldownEntry` or the map value so each key can point to its list element.
2. Maintain a list ordered from oldest to newest failure.
3. On `markFailure`:
   - Insert new keys at the newest side.
   - Move existing keys to the newest side when refreshed.
   - Evict from the oldest side when capacity is exceeded.
4. On delete/sweep, remove the map entry and the list element together.
5. Keep `newCooldownKey` unchanged.

## Files to inspect

- `pkg/filter/llm/proxy/filter.go`
- `pkg/filter/llm/proxy/filter_test.go`

## Out of scope

- Do not change cooldown TTL semantics.
- Do not change endpoint identity key fields.
- Do not replace the store with a third-party cache dependency.
- Do not add a background goroutine.

## Acceptance criteria

- Eviction no longer scans all entries to find the oldest key.
- Tests cover:
  - Oldest entry is evicted at capacity.
  - Refreshing an existing entry updates its recency.
  - TTL sweep removes both map and list state.
  - Deleting an expired cooldown does not leave a stale list element.
- Existing cooldown behavior remains unchanged from the caller's perspective.
- Tests pass:

```bash
go test ./pkg/filter/llm/proxy/...
```

## Difficulty

Medium. This is a good community task for someone comfortable with Go maps, `container/list`, and unit tests. The scope is nicely contained.

mochengqian:
## Summary

Closes #943.

The LLM proxy cooldown store is bounded by `maxCooldownStoreEntries`. At capacity it evicted the oldest entry by scanning the whole map to find the minimum `lastFailure` — O(N) on the failure path. This PR replaces that scan with a standard-library `container/list` LRU so eviction is O(1), while preserving the existing key and TTL behavior.

## Change

- `cooldownStore` now holds a `recencyOrder *list.List` (oldest at front, newest at back) and the map is keyed to `*list.Element`.
- `cooldownEntry` carries its own `key`, so a front-pop eviction can delete the matching map entry without a reverse lookup.
- `markFailure`: existing key → update entry + `MoveToBack`; new key → `evictOldestIfFullLocked` then `PushBack`.
- `evictOldestIfFullLocked`: pops `recencyOrder.Front()` in O(1) (was a full map scan).
- A shared `removeLocked(key, element)` helper keeps the map and list in sync on every delete / sweep / eviction, so no stale list element survives.

Caller-visible cooldown semantics are unchanged: endpoint identity key (`newCooldownKey` untouched), TTL derivation, idempotent replay, and the lazy expiry sweep all behave exactly as before. The one deliberate difference is the eviction victim: the old scan dropped the entry with the smallest `lastFailure` timestamp, while the LRU drops the recency-list front (insertion / refresh order). On the production failure path these always coincide, because `markFailure` is only ever called with a monotonic `time.Now()`; they can diverge only under out-of-order timestamps supplied directly in tests. No third-party dependency, no background goroutine.

## Tests

Existing cooldown tests pass unchanged (eviction-at-capacity, address/credential isolation, lazy sweep across clusters). New tests cover the LRU-specific invariants the issue calls out:

- `TestCooldownStoreRefreshUpdatesRecency` — re-failing an entry moves it to the newest side; a later eviction drops a genuinely older entry.
- `TestCooldownStoreSweepRemovesMapAndListState` — the TTL sweep clears both the map and the recency list (asserts `recencyOrder.Len() == 0`).
- `TestCooldownStoreDeleteExpiredLeavesNoStaleListElement` — clearing an expired cooldown via the request path leaves no stale list element.

Each new test asserts `len(map) == recencyOrder.Len()` to pin map/list sync.

## Test plan

- [x] `go test -race ./pkg/filter/llm/proxy/...`
- [x] `go vet ./pkg/filter/llm/proxy/` clean
- [x] `go build ./...` clean
- [x] pre-push local-CI (gofmt/vet/tests) passed

### DIFF
--- a/pkg/filter/llm/proxy/filter.go
+++ b/pkg/filter/llm/proxy/filter.go
@@ -515,7 +515,7 @@ func (s *cooldownStore) markFailure(clusterName string, endpoint *model.Endpoint
 		s.recencyOrder.MoveToBack(element)
 		return
 	}
-	s.evictOldestIfFullLocked(key)
+	s.evictOldestIfFullLocked()
 	s.lastFailureByEndpoint[key] = s.recencyOrder.PushBack(&cooldownEntry{
 		key:         key,
 		lastFailure: lastFailure,
@@ -568,11 +568,8 @@ func (s *cooldownStore) sweepExpiredExceptLocked(now time.Time, current cooldown
 }
 
 // evictOldestIfFullLocked removes the least-recently-failed entry when the
-// store is at capacity, in O(1) via the front of the recency list. current is
-// the key about to be inserted; it is never in the list yet on this path, but
-// the guard keeps the invariant explicit so a future caller cannot evict the
-// entry it is in the middle of writing.
-func (s *cooldownStore) evictOldestIfFullLocked(current cooldownKey) {
+// store is at capacity, in O(1) via the front of the recency list.
+func (s *cooldownStore) evictOldestIfFullLocked() {
 	if len(s.lastFailureByEndpoint) < maxCooldownStoreEntries {
 		return
 	}
@@ -580,11 +577,7 @@ func (s *cooldownStore) evictOldestIfFullLocked(current cooldownKey) {
 	if oldest == nil {
 		return
 	}
-	oldestKey := oldest.Value.(*cooldownEntry).key
-	if oldestKey == current {
-		return
-	}
-	s.removeLocked(oldestKey, oldest)
+	s.removeLocked(oldest.Value.(*cooldownEntry).key, oldest)
 }
 
 func newCooldownKey(clusterName string, endpoint *model.Endpoint) cooldownKey {
--- a/pkg/filter/llm/proxy/filter.go
+++ b/pkg/filter/llm/proxy/filter.go
@@ -19,6 +19,7 @@ package proxy
 
 import (
 	"bytes"
+	"container/list"
 	"crypto/sha256"
 	"errors"
 	"fmt"
@@ -127,12 +128,20 @@ type (
 	}
 
 	cooldownStore struct {
-		mu                    sync.Mutex
-		lastFailureByEndpoint map[cooldownKey]cooldownEntry
-		lastSweep             time.Time
+		mu sync.Mutex
+		// lastFailureByEndpoint indexes each tracked endpoint to its node in
+		// recencyOrder, so lookups stay O(1) and the node can be repositioned
+		// or removed without scanning.
+		lastFailureByEndpoint map[cooldownKey]*list.Element
+		// recencyOrder holds *cooldownEntry values ordered oldest-at-front,
+		// newest-at-back. Eviction pops the front in O(1); refreshing an entry
+		// moves its node to the back.
+		recencyOrder *list.List
+		lastSweep    time.Time
 	}
 
 	cooldownEntry struct {
+		key         cooldownKey
 		lastFailure time.Time
 		ttl         time.Duration
 	}
@@ -464,7 +473,8 @@ func (executor *RequestExecutor) markEndpointCooldown(endpoint *model.Endpoint) 
 
 func newCooldownStore() *cooldownStore {
 	return &cooldownStore{
-		lastFailureByEndpoint: map[cooldownKey]cooldownEntry{},
+		lastFailureByEndpoint: map[cooldownKey]*list.Element{},
+		recencyOrder:          list.New(),
 	}
 }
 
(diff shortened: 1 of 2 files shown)


pkg/filter/llm/proxy/filter.go
```diff
@@ -535,39 +559,35 @@ func (s *cooldownStore) sweepExpiredIfNeededLocked(now time.Time, current cooldo
...
+	if oldest == nil {
+		return
 	}
-	if found {
-		delete(s.lastFailureByEndpoint, oldestKey)
+	oldestKey := oldest.Value.(*cooldownEntry).key
+	if oldestKey == current {
+		return
 	}
+	s.removeLocked(oldestKey, oldest)
```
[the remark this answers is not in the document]
mochengqian: Good catch — fixed in 55f27151. `evictOldestIfFullLocked` is only ever reached on the brand-new-key path (the existing-key branch returns early), so `oldestKey == current` can never be true. And had it ever fired, it would have skipped eviction while `markFailure` still `PushBack`s, pushing the store to `maxCooldownStoreEntries + 1` — i.e. the guard weakened the very invariant it claimed to protect. Dropped the guard and the now-unused `current` parameter; capacity test still passes.

pkg/filter/llm/proxy/filter.go
```diff
@@ -500,13 +510,20 @@ func (s *cooldownStore) markFailure(clusterName string, endpoint *model.Endpoint
...
+	if element, ok := s.lastFailureByEndpoint[key]; ok {
+		entry := element.Value.(*cooldownEntry)
+		entry.lastFailure = lastFailure
+		entry.ttl = ttl
+		s.recencyOrder.MoveToBack(element)
+		return
 	}
-	s.lastFailureByEndpoint[key] = cooldownEntry{
+	s.evictOldestIfFullLocked()
+	s.lastFailureByEndpoint[key] = s.recencyOrder.PushBack(&cooldownEntry{
```
Similarityoung, line 525: 请让生产写入在持有 s.mu 后生成 lastFailure，再按同一顺序 PushBack，避免链表顺序与用于 TTL 的时间戳顺序分离。 markEndpointCooldown 当前先计算 time.Now() 再进入这里的互斥锁，并发请求可能让较新的失败先写入、较旧的失败后写入；容量满时下一次插入会淘汰链表头的较新失败，使仍处于 60 秒 cooldown 的 endpoint 立即恢复可选。

Similarityoung (commented):
审查覆盖 cooldown 失败记录从请求并发写入、TTL 查询与删除、过期清扫到满容量淘汰的完整状态流，并检查了 map 与链表同步。 CI 的 License/Import、Lint、Unit、Integration 与 SonarCloud 均通过，但 reviewer 自建的断网隔离测试证明新淘汰顺序在乱序写入时改变了既有行为。

pkg/filter/llm/proxy/filter.go
mochengqian: 好的，已在 b08b9b56 中修复。

根本原因正如你所指出的：`time.Now()` 在锁外采样，并发写入时较新的时间戳可能先 `PushBack`，导致链表顺序与时间戳顺序错位，容量满时淘汰的是链表头（实际上是较新的失败记录），使仍在 cooldown 内的 endpoint 提前恢复可选。

修复方案：在 `cooldownStore` 中注入 `nowFn func() time.Time`（默认 `time.Now`），并在持有 `s.mu` 后调用 `s.nowFn()` 采样 `lastFailure`，保证时间戳记录与 `PushBack` 顺序在同一临界区内完成，彻底消除乱序可能。`markFailure` 签名同步去掉了 `lastFailure time.Time` 参数，调用方不再在锁外传入时间戳。测试通过 `store.nowFn` 注入受控时钟，覆盖原有所有场景。

Similarityoung:
fix ci

Similarityoung (approved):
LGTM

pkg/filter/llm/proxy/filter.go
AlexStocks: [P1] 请补一条能守住锁内取时的并发回归测试

当前实现已经把 `nowFn()` 放进 `s.mu` 临界区，生产修复本身成立；但现有测试没有制造“旧时间先采样、却晚于新时间插入链表”的交错。我在当前 exact Head 的隔离副本把 `lastFailure := s.nowFn()` 移回 `s.mu.Lock()` 前，`go test -count=10 -race ./pkg/filter/llm/proxy` 仍然全绿。随后加入一个确定性探针：阻塞第一次 `nowFn`，让第二次失败先完成，再填满容量触发淘汰；Head 连续 20 次通过，而该变异稳定保留旧记录并删除新记录。请把这个交错补进正式测试，断言容量溢出时旧 timestamp 被淘汰、新 timestamp 仍在 cooldown，避免本次并发修复以后无声回退。

PR #967, opened 2026-06-07
perf(cluster): defer Config consistent-hash rebuild to legacy pick path

### LINKED ISSUE #957: [FEATURE] Defer Config.ConsistentHash rebuild to legacy pick path
## Background

Follow-up to #932 (issue #905 step 5). Distinct from #942 — see "Relationship to #942" below.

After #932, snapshot-aware consistent-hash balancers (`RingHashing`, `MaglevHash`) read `EndpointSnapshot.HealthyConsistentHash` (lazily built from the healthy subset). But `ClusterStore.prepareClusterConfig` still eagerly rebuilds `Config.ConsistentHash.Hash` on every config mutation:

```go
func (s *ClusterStore) prepareClusterConfig(c *model.ClusterConfig) {
	s.assembleClusterEndpoints(c)
	c.CreateConsistentHash()
}
```

`prepareClusterConfig` runs on `AddCluster`, `UpdateCluster`, `SetEndpoint`, and `DeleteEndpoint`.

## Problem

`Config.ConsistentHash.Hash` is now only consumed by the **legacy** (non-snapshot) balancer path — all bundled balancers are snapshot-aware and use the snapshot's own healthy hash. So for the common path this rebuild is dead work. For a Maglev cluster with a large table (e.g. `MaglevTableSize` in the tens of thousands), rebuilding on every endpoint add/remove is milliseconds of wasted CPU, and high-churn service-discovery environments (Nacos instances flapping) trigger it repeatedly.

## Goal

Stop rebuilding `Config.ConsistentHash.Hash` eagerly. Build it lazily, only when a legacy balancer first needs it.

## Suggested approach

1. Remove `c.CreateConsistentHash()` from `prepareClusterConfig`.
2. Add a lazy guard on `ClusterConfig`, e.g. `consistentHashOnce sync.Once` (tagged `yaml:"-" json:"-"`). **Must be a value, not a pointer** — `ClusterConfig` is deep-copied via `CloneStore` (yaml round-trip), and a pointer `Once` would be shared across stores. A fresh store should rebuild independently.
3. In the legacy branch of `pickEndpoint` (`pkg/cluster/loadbalancer/load_balancer.go`), trigger `context.Config.consistentHashOnce.Do(context.Config.CreateConsistentHash)` before using the hash.
4. The existing in-`Handler` nil fallback in `Maglev.Handler` / `RingHashing.Handler` (build-on-the-fly when `Hash == nil`) stays as a safety net.

## Relationship to #942

These are complementary, not duplicates:

- **#942** reuses the *snapshot's* `consistentHash` across publications when the healthy set is unchanged (avoids rebuild on health flaps inside the snapshot path).
- **This issue** stops `prepareClusterConfig` from eagerly building the *Config-level* `ConsistentHash.Hash` that only the legacy path reads.

Different fields, different code paths, different triggers.

## Risk

If any code reads `c.ConsistentHash.Hash` directly *before* a balancer `Handler` runs, it now sees `nil`. `grep -rn "ConsistentHash\.Hash" pkg/` to confirm; currently only `maglev_hash.go` and `ring_hash.go` read it, both inside `Handler`. Any external/xDS reader must be reviewed.

## Test plan

- [ ] New test: repeated `SetEndpoint` triggers zero `CreateConsistentHash` calls on the snapshot path (mock `ConsistentHashInitFunc` and count invocations).
- [ ] Legacy `Handler` builds the hash exactly once via the `sync.Once`.
- [ ] `BenchmarkSetEndpoint` on a large-Maglev cluster, before/after.
- [ ] `go test -race ./pkg/model/... ./pkg/server/... ./pkg/cluster/loadbalancer/...`

## Files

- `pkg/model/cluster.go`
- `pkg/cluster/loadbalancer/load_balancer.go`

mochengqian:
## Summary

Closes #957.

`ClusterStore.prepareClusterConfig` rebuilt `Config.ConsistentHash.Hash` eagerly on every `AddCluster` / `UpdateCluster` / `SetEndpoint` / `DeleteEndpoint`. After #932, that field is only consumed by the **legacy** (non-snapshot) balancer path — all bundled balancers are snapshot-aware and read the snapshot's own healthy hash (`EndpointSnapshot.HealthyConsistentHash`). So the eager rebuild is dead work on the common path, and for a large Maglev table it is **milliseconds** of wasted CPU on every endpoint churn (e.g. Nacos instances flapping).

This PR builds the Config-level hash lazily, only when a legacy balancer first needs it.

## Change

- `prepareClusterConfig` invalidates the hash (`c.ConsistentHash.Hash = nil`) instead of rebuilding it. Setting it to nil keeps the legacy path correct after endpoint changes — the next legacy pick rebuilds from current endpoints rather than serving a stale ring.
- New `ClusterConfig.EnsureConsistentHash()` builds the hash on first use (nil-check) and reuses it after.
- The legacy branch of `pickEndpoint` calls `EnsureConsistentHash()` under the existing `legacyPickMu`, before the `config := *context.Config` copy. The in-`Handler` nil fallback in Maglev/RingHash stays as a safety net.

### Reachability note

In-tree, the legacy branch that calls `EnsureConsistentHash` is currently **unreached**: every bundled balancer (Maglev, RingHash, RoundRobin, Rand, WeightRandom) implements `HandlerWithSnapshot` and is dispatched on the snapshot path. The win is therefore entirely from dropping the eager rebuild in `prepareClusterConfig`. `EnsureConsistentHash` exists so an out-of-tree balancer that only implements the legacy `Handler` still observes a non-nil `Config.ConsistentHash.Hash`, preserving the pre-#932 contract.

### Why a nil-check, not `sync.Once`

The issue suggested a value `consistentHashOnce sync.Once` field. That trips `go vet` copylocks at the legacy path's existing `config := *context.Config` shallow copy (verified). Since the legacy path is already serialized by `legacyPickMu`, a plain nil-check is sufficient and vet-clean. A fresh `ClusterConfig` from `CloneStore` (whose `Hash` interface does not survive the yaml round-trip, so it deserializes to nil) rebuilds independently on its first legacy pick — the same fresh-store-rebuilds-independently property the issue asked for.

## Risk

Per the issue: only code that reads `c.ConsistentHash.Hash` *before* a balancer `Handler` runs would now see nil. Confirmed the only non-test readers are `maglev_hash.go` / `ring_hash.go` (both inside `Handler`, both with a build-on-the-fly nil fallback) and `ConsistentHashForHealthyEndpoints` (snapshot path, has its own fallback). No xDS / adapter / configcenter / hotreload code reads the field.

## Benchmark

`BenchmarkSetEndpoint`, 64-endpoint Maglev cluster, `MaglevTableSize=65537`, metadata-only re-registration mirroring discovery churn. Each update flips between two metadata generations so every `SetEndpoint` is a genuine in-place replace that hits `prepareClusterConfig` — i.e. the benchmark exercises the rebuild path on **every** iteration rather than collapsing into the content-equal idempotent fast path.

`benchstat`, `-count=8`, Apple M5 / go1.26:

| | before (eager) | after (deferred) | delta |
|---|---|---|---|
| sec/op | 10505 µs | 53.12 µs | **−99.5%** (p=0.000) |
| B/op | 18016 KiB | 90.21 KiB | **−99.5%** (p=0.000) |
| allocs/op | 864 | 614 | −28.9% (p=0.000) |

The before case repopulates the whole 65537-slot Maglev table per mutation (~10.5 ms); the after case skips it. The win scales with `MaglevTableSize`.

> Note: an earlier revision of this benchmark reused a single update set, so after the first cycle every `SetEndpoint` short-circuited to the content-equal idempotent path and **neither** before nor after actually rebuilt the table — both measured the same ~410 ns fast path. The numbers above are from the corrected benchmark, which alternates metadata generations to force the replace path on every call.

## Test plan

- [x] New `TestClusterManager_SetEndpointDefersConsistentHashRebuild` — SetEndpoint churn triggers **zero** hash builds (counting `ConsistentHashInitFunc`); the legacy path's `EnsureConsistentHash` builds **exactly once** across repeated calls.
- [x] Updated the two existing consistent-hash assertions (`SetEndpoint...RebuildsConsistentHash`, `DeleteEndpoint...RepairsRuntimeAndConsistentHash`) to trigger the lazy build before inspecting `Config.ConsistentHash.Hash` directly.
- [x] `go test -race ./pkg/model/... ./pkg/server/... ./pkg/cluster/loadbalancer/...`
- [x] `go vet` clean on all three packages (confirms no copylocks regression)
- [x] pre-push local-CI passed


AlexStocks (approved):

AlexStocks (approved):

Similarityoung:
please resolve conflicts

mochengqian:
> please resolve conflicts

resolved

Similarityoung (approved):
LGTM

Issue #949, opened 2026-05-29
RFC: treat duplicate endpoint IDs as a hard configuration error


mochengqian:
> ⚠️ **This is an RFC for design discussion, not an implementation task.** Please do not open a PR that simply changes the warning to `panic` — the migration path is what this issue exists to align on. The issue will not be labeled `contribute welcome` until a direction is agreed.

## Background

Surfaced in #932 (review thread starting at https://github.com/apache/dubbo-go-pixiu/pull/932#issuecomment-4566120689). Reviewer's framing: *"ID is an identity, not a name; identity collisions should fail, not auto-rename. Neither static config nor dynamic registration should support duplicate IDs — I can't think of a scenario where duplicate IDs are legitimate."*

## Current behavior

Endpoint ID collisions are handled differently across three paths today:

| Path | Behavior on collision |
|---|---|
| Static YAML assembly (`assembleClusterEndpoints`) | Warn + suffix the second entry as `<id>-2` |
| Dynamic `SetEndpoint`, explicit ID | Replace the slot in place (registry-update semantics, fixed in #932) |
| Dynamic `SetEndpoint`, empty ID + hash collision + different content | Warn + suffix as `<hash>-2` |

The argument for hard-fail: under all three paths, a duplicate ID is almost certainly an operator mistake or an adapter bug. The current behavior (suffix or replace) silently keeps the cluster running with state the operator did not intend.

## Proposal direction (to be discussed, not committed to)

Phase the change over one minor-release deprecation window:

1. **Today → v1.x.next** — Keep current behavior. Add a `pixiu_endpoint_id_collision_total{cluster, source}` counter so operators can see if their config is affected. Also log at WARN with a deprecation note pointing at this issue.
2. **v1.x.next → v1.(x+1)** — Introduce `cluster.strict_unique_ids: true` opt-in flag. When set, duplicate IDs fail cluster initialization (static) or are rejected by `SetEndpoint` (dynamic). Default remains current behavior.
3. **v1.(x+2)** — Flip the default to `true`. Provide `strict_unique_ids: false` as an escape hatch for one more minor version, then remove.

## Open questions

- Does the static-YAML path actually have legitimate use cases for duplicate IDs that we should preserve? (Sister deployments, blue/green, ...?)
- For the dynamic empty-ID hash collision case: does hard-fail even make sense, or should that path stay tolerant since callers aren't claiming identity?
- Are there integration tests / staging deployments that would catch a misconfig before production fail?
[...]

PR #972, opened 2026-06-07
refactor(cluster): consolidate duplicated endpoint ID -2/-3 suffix algorithm

### LINKED ISSUE #969: refactor(cluster): consolidate duplicated endpoint ID -2/-3 suffix algorithm
## Background

Follow-up from the #950 audit (https://github.com/apache/dubbo-go-pixiu/issues/950#issuecomment-4642984375).

Two functions implement **byte-for-byte the same** endpoint-ID collision-suffix algorithm in two packages:

- `nextStableEndpointID` — `pkg/server/cluster_manager.go` (used by `assembleClusterEndpoints` on the static/dynamic config path, and by `resolveSetEndpointSlotByHash`)
- `uniqueSnapshotEndpointID` — `pkg/cluster/cluster.go` (used by `newEndpointSnapshot` on snapshot rebuild)

Both do the same thing: resolve a base ID (operator-supplied `endpoint.ID` if set, else `model.GenerateEndpointID(clusterName, endpoint)`), return it if free, otherwise append `-2`, `-3`, ... until unique within the per-cluster dedup set.

## Problem

They are kept in sync only by convention. Endpoint ID is the runtime health/cooldown key, so if a future change touches one and not the other (e.g. changes the suffix separator, or the collision base), config assembly and snapshot rebuild would silently assign **different** IDs to the same endpoint — splitting health/cooldown state for that endpoint across a snapshot rebuild.

There is no test pinning the two implementations to each other; each is only tested in isolation.

## Proposal

Lift the algorithm into a single shared helper, most naturally in `pkg/model` next to `GenerateEndpointID` (which both call), e.g.:

```go
// pkg/model
func StableUniqueEndpointID(clusterName string, endpoint *Endpoint, taken map[string]struct{}) string
```

Then `nextStableEndpointID` and `uniqueSnapshotEndpointID` become thin call-throughs (or are removed). This guarantees assembly and snapshot rebuild can never drift.

## Acceptance criteria

- One implementation of the `-2/-3` suffix algorithm; both call sites use it.
- Existing dedup tests in `pkg/server` and `pkg/cluster` still pass unchanged.
- A test asserts assembly and snapshot rebuild produce identical IDs for a duplicate-ID cluster (locks the two paths together).

## Out of scope

- Changing the dedup *semantics* (suffix-vs-replace, hard-fail on duplicates) — that is #949's discussion, not this refactor.

## Difficulty

Low. Pure consolidation, no behavior change.

mochengqian:
## Summary

Closes #969.

`nextStableEndpointID` (`pkg/server/cluster_manager.go`) and `uniqueSnapshotEndpointID` (`pkg/cluster/cluster.go`) were **byte-for-byte identical** implementations of the endpoint-ID collision-suffix algorithm, kept in sync only by convention. Since the endpoint ID is the runtime health/cooldown key, a future edit to one and not the other would silently assign **different** IDs to the same endpoint — splitting its health/cooldown state across a snapshot rebuild.

This PR lifts the algorithm into a single shared helper and routes both paths through it, so they can never drift.

## Changes

- Add `model.StableUniqueEndpointID(clusterName, endpoint, taken)` in `pkg/model/cluster.go`, next to `GenerateEndpointID` (which both call sites already use).
- Remove `nextStableEndpointID` from `pkg/server`; repoint its three call sites (`assembleClusterEndpoints` ×2, `resolveSetEndpointSlotByHash`) to the shared helper.
- Remove `uniqueSnapshotEndpointID` from `pkg/cluster`; repoint `newEndpointSnapshot` to the shared helper, and drop the now-unused `fmt` import.
- Add `TestEndpointIDAssemblyAndSnapshotRebuildAgree`: feeds the same duplicate-ID config independently through config assembly and snapshot rebuild and asserts identical ID sequences — locking the two paths together. Covers both the explicit-duplicate-ID branch (`foo` / `foo-2`) and the anonymous identical-hash branch (`generated-<hash>` / `generated-<hash>-2`).

Pure consolidation — no behavior change. The dedup *semantics* (suffix-vs-replace, hard-fail on duplicates) are out of scope, per #969 (that's #949's discussion).

## Test plan

- [x] `go test ./pkg/model/... ./pkg/server/... ./pkg/cluster/...` passes (existing dedup tests in `pkg/server` and `pkg/cluster` unchanged).
- [x] New `TestEndpointIDAssemblyAndSnapshotRebuildAgree` passes (both subtests).
- [x] `go build ./...` clean.
- [x] `gofmt -l` clean; `go vet` clean.

### DIFF
--- a/pkg/cluster/cluster.go
+++ b/pkg/cluster/cluster.go
@@ -18,7 +18,6 @@
 package cluster
 
 import (
-	"fmt"
 	"sync"
 	"sync/atomic"
 )
@@ -242,7 +241,7 @@ func newEndpointSnapshot(config *model.ClusterConfig, previous *EndpointSnapshot
 			continue
 		}
 		snapshotEndpoint := model.CloneEndpoint(endpoint)
-		snapshotEndpoint.ID = uniqueSnapshotEndpointID(clusterName, snapshotEndpoint, endpointIDs)
+		snapshotEndpoint.ID = model.StableUniqueEndpointID(clusterName, snapshotEndpoint, endpointIDs)
 		endpointIDs[snapshotEndpoint.ID] = struct{}{}
 		address := snapshotEndpoint.Address.GetAddress()
 		healthy := endpointSnapshotHealth(snapshotEndpoint, address, previous, inheritRuntimeHealth)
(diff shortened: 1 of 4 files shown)


AlexStocks (approved):

Similarityoung (approved):
LGTM

PR #973, opened 2026-06-09
feat(cluster): decouple config from runtime — private Config, RuntimeState, and RoundRobin cursor migration

### LINKED ISSUE #958: [FEATURE] Cluster config/runtime decoupling series
## Background

Follow-up to issue #905 and PR #932.

PR #932 (issue #905 step 5) introduced `EndpointSnapshot` and completed **state isolation**: runtime health changes and membership updates now go through immutable snapshots published via CAS, and the request path no longer mutates `Endpoint.UnHealthy` or reads stale `Config.Endpoints`.

However, **structural coupling** between configuration and runtime remains. `Cluster.Config` is still a public field; `ClusterStore.Config` and `clustersMap[name].Config` share the same pointer; and runtime state (e.g. `PrePickEndpointIndex`) lives on the configuration struct. PR #932 explicitly scoped these out:

> Out of scope (tracked as tech debt): `Cluster.Config` remains public in this PR. Making it private + adding getters touches ~20 callsites and belongs in the decoupling series after PR-3 lands; target: decoupling PR-6, together with runtime ownership cleanup.

This issue tracks that decoupling series.

## Current coupling points

1. **`Cluster.Config` is a public field** (`pkg/cluster/cluster.go:40`) — external code can directly read/write `ClusterConfig`, bypassing any encapsulation or validation. No getter/setter boundary.

2. **`ClusterStore.Config` and runtime `Cluster.Config` share the same pointer** — `replaceClusterRuntime` assigns `config` directly to the new cluster (`pkg/server/cluster_manager.go:448`). Mutations to `store.Config[i]` are visible to the runtime cluster, and vice versa. The identity check `runtimeCluster.Config != clusterConfig` (line 475) relies on pointer equality, which is fragile under deep-copy or reload scenarios.

3. **RoundRobin cursor lives on `ClusterConfig.PrePickEndpointIndex`** — configuration structs represent desired state; runtime pick cursors are operational state. The two should not share storage. The current design forces legacy and snapshot RoundRobin paths to contend on the same atomic counter (see `pkg/cluster/loadbalancer/load_balancer.go` RR fairness caveat comment), creating a subtle fairness skew risk.

4. **`prepareClusterConfig` eagerly rebuilds `Config.ConsistentHash.Hash`** on every endpoint mutation (`pkg/server/cluster_manager.go:362`), even though snapshot-aware balancers read `EndpointSnapshot.HealthyConsistentHash` and never touch `Config.ConsistentHash.Hash`. This is wasted work and should be deferred to the legacy pick path (tracked separately in #957, but the root cause is that config and runtime hash storage are conflated).

5. **Ownership boundaries unclear** — when `ClusterStore` calls `replaceClusterRuntime`, the old runtime is returned and stopped by the caller. But there's no clear contract on who owns the `ClusterConfig` lifecycle, or when it's safe to mutate `Config.Endpoints` vs. when the runtime has published a snapshot from it.

## Goal

Establish a clean separation:

- **Configuration layer** (`ClusterStore.Config`, `ClusterConfig`) represents desired state. Immutable after publication to runtime. External mutations go through `ClusterManager` APIs that trigger controlled runtime updates.
- **Runtime layer** (`clustersMap[name]`, `Cluster`, `EndpointSnapshot`) represents operational state. Holds its own copy (or read-only view) of config. Runtime state (cursors, health, snapshots) never leaks into config structs.
- **Clear ownership** — who owns each object, when it's safe to mutate, and how lifecycle (start/stop/replace) is managed.

## Suggested approach (sequenced steps)

This is a **series**, not a single PR. Each step should be independently reviewable and have clear before/after semantics.

### Step 1: Make `Cluster.Config` private + add getters

- Rename `Cluster.Config` → `Cluster.config` (unexported).
- Add `func (c *Cluster) Config() *model.ClusterConfig` that returns a **read-only view** (either a shallow copy with critical fields frozen, or a wrapper that panics on write — TBD during implementation).
- Audit all ~20 external read sites and replace `cluster.Config.X` with `cluster.Config().X`.
- For the few write sites (if any exist outside `ClusterManager`), route them through explicit `ClusterManager` mutation APIs or flag them as bugs.

**Goal:** No external code can directly mutate `Cluster.config` after it's constructed.

### Step 2: `replaceClusterRuntime` deep-clones `ClusterConfig`

- Instead of `s.clustersMap[name] = cluster.NewClusterWithEndpointSnapshot(config, previous)` (shared pointer), deep-clone `config` first:
  ```go
  runtimeConfig := model.CloneClusterConfig(config)  // new helper
  s.clustersMap[name] = cluster.NewClusterWithEndpointSnapshot(runtimeConfig, previous)
  ```
- The runtime cluster now owns its config copy. Mutations to `store.Config[i]` no longer affect the running cluster until the next `replaceClusterRuntime`.
- The identity check `runtimeCluster.Config != clusterConfig` becomes a generation/version check instead of pointer equality (consider adding `ClusterConfig.Generation int64` or similar).

**Goal:** `store.Config` and `clustersMap[name].Config()` point to independent objects.

### Step 3: Move RoundRobin cursor out of `ClusterConfig`

- Add a `RuntimeState` struct (or similar) that holds operational counters:
  ```go
  type RuntimeState struct {
      roundRobinCursor uint32
  }
  ```
- `Cluster` holds `runtimeState *RuntimeState`.
- `PickContext` carries `runtimeState` alongside `Config` and `Snapshot`.
- RoundRobin reads/writes the cursor from `runtimeState`, not `Config`.
- Legacy RoundRobin path and snapshot RoundRobin path now have **independent cursors** (or share one in `runtimeState` if that's the desired semantic — but either way, it's not on the config struct).

**Goal:** `ClusterConfig` is pure desired-state data. No runtime operational fields.

### Step 4: Audit and document ownership boundaries

- Write down the lifecycle contract:
  - Who creates `ClusterConfig`? (Answer: user yaml / xDS / `ClusterManager.AddCluster`)
  - When is it safe to mutate? (Answer: only before passing to `AddCluster` / `UpdateCluster`, or inside `prepareClusterConfig` before runtime publication)
  - Who owns runtime `Cluster`? (Answer: `ClusterStore.clustersMap`, lifecycle managed by `replaceClusterRuntime` / `stopClusters`)
  - When is a `Cluster` stopped? (Answer: on `UpdateCluster` / `DeleteCluster` / `CompareAndSetStore`)
- Add godoc on `ClusterConfig`, `Cluster`, `ClusterStore` summarizing the contract.
- Add a test that tries to mutate `store.Config[i].Endpoints` after `AddCluster` and asserts the runtime cluster doesn't see it (proving the clone).

**Goal:** Clear ownership rules documented and enforced by tests.

### Step 5 (optional stretch): Immutable `ClusterConfig` with builder pattern

If steps 1-4 reveal lots of subtle mutation sites, consider:

- `ClusterConfig` becomes append-only (all fields have setters that return a new copy).
- Or use a builder pattern (`ClusterConfigBuilder`) that produces immutable `ClusterConfig` on `.Build()`.

This is a bigger refactor and may not be worth it if steps 1-4 already establish clear boundaries. Defer unless mutation bugs keep surfacing.

## Acceptance criteria

- `Cluster.Config` is unexported; external reads go through a getter that returns a read-only view.
- `store.Config[i]` and `clustersMap[name].Config()` point to independent objects (verified by pointer inequality test).
- RoundRobin cursor is in runtime state, not in `ClusterConfig`.
- Ownership contract is documented in godoc and enforced by tests.
- Mutating `store.Config[i]` after `AddCluster` / `UpdateCluster` has no effect on the running cluster until the next explicit update API call.

## Relationship to other follow-ups

- **Independent of** #938-950, #955-957 (those are perf / API cleanup; this is structural)
- **Complements** #957 (lazy `Config.ConsistentHash`) — once config and runtime are separate, it's clearer that `Config.ConsistentHash.Hash` is legacy-only

## Files

- `pkg/cluster/cluster.go`
- `pkg/model/cluster.go`
- `pkg/cluster/loadbalancer/roundrobin/round_robin.go` (step 3)
- `pkg/cluster/loadbalancer/load_balancer.go` (step 3)

## Priority

Medium. State isolation is already complete (#932), so this is not a correctness blocker. But it's foundational for:

- Making the cluster runtime easier to reason about (clear mutation boundaries)
- Enabling future work like dynamic reload without restart, or separating control-plane config from data-plane runtime
- Reducing the risk of accidental config mutation bugs

Suggest tackling this after the current perf follow-ups (#955-957) land, so the architecture is stable before the next round of optimization.

Qiao-yq:
**What this PR does**:

This PR completes the config-runtime decoupling identified as technical debt in #932. It establishes a clear boundary between the configuration layer (ClusterConfig) and the runtime layer (Cluster), ensuring that mutations in one do not leak into the other.

**Changes**:

1. **Config identity without pointer comparison**
   - Added `ConfigID` (atomic sequence) to `ClusterConfig` for stable identity after deep copy
   - Added `ConfigIsIdenticalTo()` to replace fragile pointer comparisons
   - Added `CloneClusterConfig()` helper with independent copies of Endpoints, HealthChecks, and ConsistentHash

2. **Private `Cluster.config` with getter**
   - Renamed `Cluster.Config` → unexported `config` with `Config()` getter
   - Prevents external code from bypassing encapsulation

3. **RoundRobin cursor moved to runtime**
   - Added `RuntimeState` struct on `Cluster` holding `roundRobinCursor`
   - Added `RoundRobinCursor` field to `loadbalancer.PickContext`
   - Snapshot-path RoundRobin now uses the runtime-owned cursor; `PrePickEndpointIndex` is legacy fallback only

4. **Deep copy on runtime publication**
   - `replaceClusterRuntimeWithSnapshot` now deep-copies config via `CloneClusterConfig`, so store and runtime hold independent objects

Fixes #958

NONE

```release-note
NONE

### CODE REVIEW

File: pkg/cluster/loadbalancer/load_balancer.go [hunk 1]
```diff
@@ -173,10 +176,15 @@ func pickEndpoint(balancer LoadBalancer, context PickContext, policy model.LbPol
...
+		cursorBefore = context.RoundRobinCursor.Load()
+	} else {
+		cursorBefore = atomic.LoadUint32(&context.Config.PrePickEndpointIndex)
+	}
 	atomic.StoreUint32(&config.PrePickEndpointIndex, cursorBefore)
 	endpoint := balancer.Handler(&config, policy)
-	cursorAfter := atomic.LoadUint32(&config.PrePickEndpointIndex)
+	cursorAfter = atomic.LoadUint32(&config.PrePickEndpointIndex)
 	if cursorAfter != cursorBefore {
 		atomic.AddUint32(&context.Config.PrePickEndpointIndex, cursorAfter-cursorBefore)
```
AlexStocks: [P1] 当 `PickContext.RoundRobinCursor` 非 nil 时，legacy balancer 的 cursor delta 仍然回写到 `context.Config.PrePickEndpointIndex`，运行时的 `RoundRobinCursor` 没有前进。这个 PR 把正常请求路径的游标迁到了 `Cluster.runtimeState`，`ClusterManager` 也会传入 `RoundRobinCursor`；因此仍走旧 `Handler(*ClusterConfig, Policy)` 的自定义负载均衡器如果依赖 `PrePickEndpointIndex`，下一次 pick 仍会从旧 runtime cursor 开始，轮询状态丢失。这里应在 `RoundRobinCursor` 非 nil 时把 delta 写回 `context.RoundRobinCursor`，只有 nil 时才回写 `Config.PrePickEndpointIndex`。

AlexStocks (approved):

Similarityoung (commented):
please resolve conflict

Qiao-yq:
> please resolve conflict

非常抱歉qwq刚刚才看到

pkg/model/cluster.go
```diff
@@ -305,3 +315,39 @@ func cloneLLMMeta(meta *LLMMeta) *LLMMeta {
...
+	}
+	clone := *c
+	clone.Endpoints = CloneEndpoints(c.Endpoints)
+	clone.HealthChecks = cloneHealthChecks(c.HealthChecks)
+	clone.ConsistentHash = ConsistentHash{
+		ReplicaNum:      c.ConsistentHash.ReplicaNum,
+		MaxVnodeNum:     c.ConsistentHash.MaxVnodeNum,
+		MaglevTableSize: c.ConsistentHash.MaglevTableSize,
+		Hash:            nil,
+	}
```
Similarityoung: 当 `LbStr` 没有注册工厂时，请保留 `ConsistentHash.Hash`。 `prepareOwnedClusterConfig` 明确支持为此类自定义策略注入哈希，但这里在发布运行时前始终把它替换为 nil，导致读取该哈希的已注册快照负载均衡器无法选出端点。 审查测试 `TestReviewerRuntimePreservesProgrammaticCustomHash` 在 base 通过、在当前 head 因选中端点为 nil 而失败，因此应只在哈希能够重建时清空它，否则保留调用方注入的实现。

Similarityoung (commented):
审查了配置克隆与身份、运行时快照发布、端点与健康检查同步，以及 RoundRobin 游标迁移和继承。 CI 全部通过，但审查者编写的聚焦测试证明，运行时发布会丢弃无工厂自定义负载均衡策略由程序注入的哈希；该测试在 base 通过、在 head 失败。

Similarityoung (approved):
LGTM

PR #981, opened 2026-06-16
refactor: replace goinggo/mapstructure with mitchellh/mapstructure

Aias00:
## Summary

Replace the obsolete `github.com/goinggo/mapstructure` fork (v0.0.0-20140717182941, from 2014) with the standard `github.com/mitchellh/mapstructure` v1.5.0, which is already a dependency in go.mod.

## Changes

- `pkg/config/config_load.go`: Updated import from `github.com/goinggo/mapstructure` → `github.com/mitchellh/mapstructure`
- `go.mod`: Removed `github.com/goinggo/mapstructure` dependency
- `go.sum`: Removed checksum entries for the old fork

## Why

- `goinggo/mapstructure` is a 2014 fork with no updates in 12 years
- It was only used in one place (`config_load.go`) for `mapstructure.Decode()`
- The standard `mitchellh/mapstructure` v1.5.0 is already in go.mod and used by 6 other files in the codebase with the same API
- Removing the fork reduces dependency count and maintenance burden

## Verification

- ✅ `go build ./...` passes
- ✅ `go test ./pkg/config/...` passes
- ✅ `go mod tidy` removes the old dependency cleanly

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -25,7 +25,6 @@ require (
 	github.com/go-playground/assert/v2 v2.2.0
 	github.com/go-resty/resty/v2 v2.7.0
 	github.com/go-sql-driver/mysql v1.9.2
-	github.com/goinggo/mapstructure v0.0.0-20140717182941-194205d9b4a9
 	github.com/golang-jwt/jwt/v4 v4.5.2
 	github.com/golang/mock v1.6.0
 	github.com/golang/protobuf v1.5.4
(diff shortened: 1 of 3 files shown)


Aias00:
Replying to @Copilot's review comment on `go.mod`:

Thanks for the review! This is actually a false positive — `github.com/mitchellh/mapstructure v1.5.0` is already present in `go.mod` (line 37) as a direct dependency, used by 6 other files in the codebase (`pkg/model/http.go`, `pkg/model/listener.go`, `pkg/config/xds/xds.go`, `pkg/tracing/jaeger/jaeger.go`, `pkg/cluster/retry/exponentialbackoff/exponential_back_off.go`, `pkg/wasm/service.go`). We only replaced the import path in `config_load.go` from the obsolete fork to the existing standard package, so no new dependency entry is needed. `go mod tidy`, `go build ./...`, and `go test ./pkg/config/...` all pass cleanly.

Similarityoung (approved):
lgtm

Issue #989, opened 2026-06-25
[BUG] Review findings on develop: startup lifecycle, Kafka facade panic, and aggressive shutdown/error handling


AlexStocks:
## Summary

I reviewed the current `develop` branch of `apache/dubbo-go-pixiu` locally and found several source-grounded stability / lifecycle issues that are worth follow-up.

This issue is intentionally focused on problems that are either:
- obvious runtime bugs,
- goroutine / lifecycle / shutdown hazards,
- or panic-style failure handling that is too aggressive for ordinary runtime/configuration errors.

---

## 1. `KafkaConsumerFacade.Subscribe()` can panic on first use because `consumerManager` is never initialized

**File**: `pkg/client/mq/kafka_facade.go`

Relevant code path:
- `KafkaConsumerFacade` declares `consumerManager map[string]func()`
- `NewKafkaConsumerFacade(...)` initializes `consumerGroup`, `httpClient`, `done`, but does **not** initialize `consumerManager`
- `Subscribe()` does:

```go
key := GetConsumerManagerKey(cOpt.TopicList, cOpt.ConsumerGroup)
f.consumerManager[key] = cancel
```

If `consumerManager` is still nil, that assignment panics immediately.

### Why this matters

This is not a theoretical issue: a nil map write in Go is a direct runtime panic.

### Suggested fix

- initialize `consumerManager` in `NewKafkaConsumerFacade(...)`, e.g. `make(map[string]func())`
- also review concurrent access to `consumerManager`, because it is later read/written from background goroutines without synchronization

### Suggested follow-up test

- add a unit test that constructs a new consumer facade and calls `Subscribe()` once
- expected behavior: no panic

---

## 2. Pixiu startup wait group appears permanently unbalanced, which can block startup forever

**File**: `pkg/server/pixiu_start.go`

Relevant code path:

```go
func (s *Server) Start() {
    ...
    s.startWG.Add(1)
    ...
}

func Start(bs *model.Bootstrap) {
    ...
    server.Start()
    server.startWG.Wait()
}
```

I did not find a matching `Done()` for this `startWG.Add(1)` in the startup path.

### Why this matters

If the wait group counter is incremented and never decremented, `server.startWG.Wait()` will block indefinitely.

This looks like a straightforward lifecycle bug unless there is an external `Done()` path that I missed.

### Suggested fix

- either remove this wait group if it is no longer needed,
- or ensure the intended startup completion path calls `Done()` exactly once
[...]

PR #991, opened 2026-06-26
fix(server): remove unbalanced startup wait group

Aias00:
## Summary

The `startWG sync.WaitGroup` in `Server` struct had `Add(1)` called in `Start()` but **no matching `Done()` call anywhere** in the codebase. This made `server.startWG.Wait()` block indefinitely by accident — a misleading construct that appeared to coordinate something but was really just an unbalanced WaitGroup.

## Change

- Removed `startWG sync.WaitGroup` field from `Server` struct
- Removed `s.startWG.Add(1)` from `Server.Start()`
- Replaced `server.startWG.Wait()` with `select {}` — the Go idiom for "block forever"
- Removed unused `sync` import

## Why `select {}` is correct

The application shuts down via signal handling in `listener_manager.go:gracefulShutdownInit()`, where **every exit path calls `os.Exit(0)`**:

| Path | Line |
|------|------|
| Timeout during graceful shutdown | L103 |
| Error from listener `ShutDown()` | L112 |
| Normal completion after all listeners shut down | L124 |

There is no channel, context, or callback to notify the main goroutine of shutdown — the process simply exits. `select {}` is functionally equivalent to the old behavior but makes the intent explicit.

No tests are affected; no other code referenced `startWG`.

---

### DIFF
--- a/pkg/server/pixiu_start.go
+++ b/pkg/server/pixiu_start.go
@@ -20,7 +20,6 @@ package server
 import (
 	"net/http"
 	"strconv"
-	"sync"
 )
 
 import (
@@ -35,8 +34,6 @@ var server *Server
 
 // PX is Pixiu start struct
 type Server struct {
-	startWG sync.WaitGroup
-
 	listenerManager *ListenerManager
 	clusterManager  *ClusterManager
 	adapterManager  *AdapterManager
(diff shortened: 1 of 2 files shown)


Aias00:
link: https://github.com/apache/dubbo-go-pixiu/issues/989 Pixiu startup wait group appears permanently unbalanced, which can block startup forever

AlexStocks (approved):

AlexStocks (approved):

Similarityoung (approved):
LGTM

PR #993, opened 2026-06-26
fix(server): improve listener shutdown error handling

### LINKED ISSUE #989: [BUG] Review findings on develop: startup lifecycle, Kafka facade panic, and aggressive shutdown/error handling

Aias00:
## Summary

Fixes #989 Issue 3 (listener/shutdown path): The original implementation called `os.Exit(0)` from worker goroutines during shutdown, which is too aggressive.

### Original Problems

- `os.Exit(0)` was called in per-listener worker goroutines on shutdown error
- `os.Exit(0)` was called in timeout handler (`time.AfterFunc`)
- Process always exited with success code (0), even when shutdown errors occurred
- One listener shutdown failure would immediately exit the process while other cleanup was still in progress

### Why this matters

These are operational stability issues:
- Configuration mistakes or transient dependency failures can terminate the whole process
- One listener shutdown failure can immediately exit the process while other cleanup is still in progress
- Metrics/tracing/log flush may not complete
- Makes the system harder to embed, test, and recover

### Changes

- Remove `os.Exit(0)` calls from per-listener worker goroutines
- Remove `os.Exit(0)` from timeout handler (`time.AfterFunc`)
- Collect shutdown errors in a buffered channel instead of immediate exit
- Wait for all shutdowns to complete or timeout before proceeding
- Exit with appropriate code: 0 for success, 1 if errors occurred
- Add unit tests for error collection, timeout handling, and WaitGroup behavior

### Note on WaitGroup

`listener.ShutDown()` methods internally call `wg.Done()`, so we do not add `defer wg.Done()` in the worker goroutine to avoid double `Done()` calls which would panic with negative WaitGroup counter.

## Test Plan

- Added `pkg/server/listener_manager_test.go` with tests covering:
  - WaitGroup double-Done panic behavior
  - Shutdown error collection pattern
  - Shutdown timeout handling pattern
  - Labeled break pattern for draining channels

### DIFF
--- a/pkg/server/listener_manager.go
+++ b/pkg/server/listener_manager.go
@@ -46,6 +46,10 @@ type wrapListenerService struct {
 	config *model.Listener
 }
 
+// ShutdownFunc is the signature for a listener shutdown function.
+// It returns an error if shutdown failed.
+type ShutdownFunc func() error
+
 // ListenerManager the listener manager
 type ListenerManager struct {
 	bootstrap *model.Bootstrap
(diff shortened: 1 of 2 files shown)


pkg/server/listener_manager.go
```diff
@@ -98,33 +102,124 @@ func (lm *ListenerManager) gracefulShutdownInit() {
...
+	// We use doneCh instead of relying solely on WaitGroup because:
+	// - listener.ShutDown() calls wg.Done() before returning
+	// - this ensures we wait until error is sent to errCh
+	allDone := make(chan struct{})
+	go func() {
+		for i := 0; i < len(shutdownFuncs); i++ {
+			select {
+			case <-doneCh:
+				// listener goroutine completed
+			case <-time.After(perListenerTimeout):
```
AlexStocks: [P0] 这里额外引入固定 5 秒的 `perListenerTimeout`，会覆盖用户配置的整体优雅关闭语义。现有 TCP/HTTP2/Triple/gRPC listener 都可能按 bootstrap 中的 shutdown timeout 等待活跃请求；例如配置 30 秒且只有一个 listener 时，本循环 5 秒后就把该 listener 当作已处理并关闭 `allDone`，随后调用方以 `timedOut=false` 执行 `os.Exit(0)`，直接中断本应继续 25 秒的优雅关闭。不要用无身份的固定超时消费“完成名额”；应只由整体配置 timeout 控制退出，或为每个 listener 使用不短于剩余整体期限的独立上下文并明确记录未完成项。

AlexStocks (changes requested):
发现 1 个 P0，详见 Files changed 行内评论。

PR #995, opened 2026-06-26
fix(admin): replace panic/os.Exit with error handling in config loading

### LINKED ISSUE #989: [BUG] Review findings on develop: startup lifecycle, Kafka facade panic, and aggressive shutdown/error handling

Aias00:
## Summary

Fixes #989 Issue 3 (admin config loading path): Replace panic/os.Exit with proper error handling in admin config loading.

### Problem

The original implementation used panic and os.Exit in multiple places:
- `viper.go`: panic on config file read error
- `xds.go`: os.Exit(1) on startup validation failures
- `xds.go`: panic/os.Exit in background goroutine on runtime failures

### Changes in viper.go
- `Viper()` now returns `(*viper.Viper, error)` instead of panic on config read error
- Update `server.go` caller to handle the error gracefully

### Changes in xds.go
- **Startup errors**: return error instead of os.Exit(1)
  - config.Consistent() failure → return error
  - snapshot.SetSnapshot() failure → return error
- **Background goroutine**: log and continue instead of exit
  - WatchWithPrefix error → log and return (keep running)
  - config reload errors → log and continue (keep previous valid config)
- **Invariant violation**: keep panic in makeHTTPFilter (true programming error)

### Classification

| Location | Type | Classification | Fix |
|----------|------|----------------|-----|
| viper.go:59 | panic | Config/startup error | Return error |
| xds.go:95,101 | os.Exit | Config/startup error | Return error |
| xds.go:151,160,166 | panic/exit | Runtime failure | Log and continue |
| xds.go:207 | panic | True invariant | Keep |

### Why This Matters

Background goroutines should not exit the process directly. They should:
- Log errors and continue running with previous valid state
- Optionally signal for graceful shutdown

## Test Plan

- Build passes: `go build ./admin/...`


admin/core/server.go
```diff
@@ -48,9 +48,13 @@ type server interface {
 }
 
 // RunServer start server
-func RunServer() {
+func RunServer() error {
 	// load config
-	global.VP = Viper()
+	vp, err := Viper()
```
AlexStocks: [P1] CLI 的 `--config/-c` 写入 `cmd/admin.configPath`，但新的错误传播链仍然调用无参数 `Viper()`，没有把该路径传入；`Viper()` 会继续走自己的默认/stdlib flag 逻辑。结果是用户显式指定的 Admin 配置文件仍可能被忽略。请把 `configPath` 通过 `Start(configPath) → RunServer(configPath) → Viper(configPath)` 明确传递，并增加非默认路径测试。

admin/core/server.go
```diff
@@ -83,4 +87,5 @@ func RunServer() {
 	})
 
 	wg.Wait()
+	return nil
```
AlexStocks: [P0] 新增的 `error` 返回实际上无法传播 HTTP/xDS 启动失败：两个 worker 在 `ListenAndServe` 或 `StartxDsServer` 出错时都只记录日志并正常返回，`wg.Wait()` 结束后这里固定返回 `nil`。因此端口占用、xDS snapshot 初始化失败等场景下，`cmd/admin` 的 `errCh` 仍会报告成功或继续等待，进程形成假健康状态。worker 必须把首个非预期错误发送到协调通道并触发其余服务关闭，`RunServer` 再将该错误返回 CLI；同时补充端口绑定失败和 xDS 初始化失败测试。

AlexStocks (changes requested):
发现 1 个 P0、1 个 P1，详见 Files changed 行内评论。

PR #997, opened 2026-06-26
fix(cmd): replace panic with error returns in gateway startup

### LINKED ISSUE #989: [BUG] Review findings on develop: startup lifecycle, Kafka facade panic, and aggressive shutdown/error handling

Aias00:
## Summary

Fixes #989 Issue 3 (gateway startup path): Replace panic with proper error returns in gateway startup command.

### Problem

The original implementation used panic in cobra command handlers:
- `PreRun`: panic when `initialize()` fails
- `Run`: panic when `start()` fails
- `stop()`: panic as placeholder for unimplemented method

Stack traces from configuration errors are not helpful to users.

### Changes
- `PreRun` → `PreRunE`: Use cobra's error handling
  - `initialize()` failure now returns wrapped error to cobra
- `Run` → `RunE`: Use cobra's error handling
  - `start()` failure now returns wrapped error to cobra
- `stop()`: Return error instead of panic
  - Changed from `panic("implement me")` to `errors.New("stop not implemented")`

### Classification

| Handler | Type | Classification | Fix |
|---------|------|----------------|-----|
| PreRun panic | Config/startup error | Return error to cobra |
| Run panic | Startup error | Return error to cobra |
| stop panic | Placeholder | Return error |

### Benefits

- Cobra handles errors gracefully, printing clean error messages
- Stack traces are no longer shown for configuration errors
- Users get clearer feedback on what went wrong
- Future changes to `server.Start()` can return errors properly

## Test Plan

- Build passes: `go build ./pkg/cmd/...`

### DIFF
--- a/pkg/cmd/gateway.go
+++ b/pkg/cmd/gateway.go
@@ -58,7 +58,7 @@ var (
 		Short: "Run dubbo go pixiu in gateway mode",
 	}
 
-	deploy = &DefaultDeployer{
+	deploy Deployer = &DefaultDeployer{
 		configManger: config.NewConfigManger(),
 	}
 
--- a/pkg/cmd/gateway.go
+++ b/pkg/cmd/gateway.go
@@ -25,6 +25,8 @@ import (
 )
 
 import (
+	"github.com/pkg/errors"
+
 	"github.com/spf13/cobra"
 )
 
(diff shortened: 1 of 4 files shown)


pkg/cmd/gateway.go
```diff
@@ -56,27 +58,28 @@ var (
 		Short: "Run dubbo go pixiu in gateway mode",
 	}
 
-	deploy = &DefaultDeployer{
+	deploy Deployer = &DefaultDeployer{
```
AlexStocks, line 61: 这个类型没必要加上

AlexStocks (approved):

Similarityoung (approved):
lgtm

